PR #2, opened 2023-11-29
add ability to choose a package manager

shreyassanthu77:
add a cli flag to choose a package manager to install deps and intelligently chose either `pnpm`, `yarn` or `npm` if available in path in that order by default

### DIFF
--- a/cmd/sst/main.go
+++ b/cmd/sst/main.go
@@ -23,6 +23,10 @@ func main() {
 			&cli.BoolFlag{
 				Name: "verbose",
 			},
+			&cli.StringFlag{
+				Name:    "package-manager",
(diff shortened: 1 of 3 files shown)


thdxr:
So we actually want to keep it npm since this should be totally invisible to the user - a single package manager keeps things more maintainable for us

at some point we're likely going to get rid of the need to use a package manager all together and bundle the dependencies into a single blob

shreyassanthu77:
Yeah, makes sense. When I use bun to install dependencies the ci times go to seconds from minutes so I thought it'd be a good option to have. But now that you mention it bundling the packages is a better idea ( but I am not sure how typescript LSP works with that tho  )

PR #6, opened 2024-01-22
fix: use us-east-1 provider when creating redirect certificate

lukeshay:
When you deploy to a region other than `us-east-1` (e.g. `us-west-2`), this certificate is created in that region (`us-west-2`). This change uses the `useProvider` hook so the certificate is created in `us-east-1` to be used by Cloudfront.


fwang:
Nice catch. Thanks @lukeshay!

PR #8, opened 2024-01-31
Open-next V3 in ion

conico974:
## WIP
A very basic implementation of Open-Next V3 in ion. 

Couple of things to note : 
- Only lambda is supported (with or without streaming)
- Can't customize functions
- No warmer
- Only combined logging
- Separate implementation from open-next V2


fwang:
Done 🎉

PR #9, opened 2024-02-02
Go build errors

jackbravo:
I was trying to build the project to test latest sst locally and ran into some errors.


thdxr:
this is no longer need - thank you though!

PR #17, opened 2024-03-08
Update curl URL

dylanrod21:
https://ion.sst.dev/install.sh returns a 403, removing the '.sh' resolves the issue


jayair:
Hmm I think somebody fixed it in the source but the docs haven't been regenerated yet.

https://github.com/sst/ion/blob/dev/cmd/sst/main.go#L181

But thanks for pointing it out, I need to fix the 403 error on those pages.

dylanrod21:
Cool thanks, I'm not familiar with go, I didn't realize the documentation is in the code, that's pretty neat.

PR #20, opened 2024-03-09
fix(docs): Minor typos

garretcharp:


garretcharp:
While `it'll` is not a typo it read weirdly to me so I changed it.

jayair:
Yeah some of these are generated. Left a link to their source, make the change there and resubmit?

Don't worry about the generated files, it'll get overwritten on deploy.

garretcharp:
Should be fixed in the places it gets generated from now @jayair

jayair:
Thanks!

PR #23, opened 2024-03-11
fix: Make sure sst_links are set for cjs bundled functions

cmanou:
- GlobalThis wasn't being set with the links if you didn't pass a banner (which I wasn't)
- Just copy the same logic as esm just above


thdxr:
thank you!

PR #24, opened 2024-03-11
fix: Grant more explicit permissions for postgres link

cmanou:
- A common use case is using drizzle or similar orms which require more than just execute statement. I opted not to use `rds-data:*` just incase if they are actions added later we will explicitly add them.


garretcharp:
ha yeah I ran into this added the perms myself on the function I needed it but would be nice if link added them

PR #28, opened 2024-03-12
nextjs: add basepath prefixPattern func

### LINKED ISSUE #4935: Support `basePath` in Nextjs
As described in [this issue](https://github.com/sst/sst/issues/2964) in sst/sst and as addressed in this sst/sst [PR](https://github.com/sst/sst/pull/3613/files), if `basePath` is set in `next.config.js`, sst/ion ignores that path when defining Cloudfront routes, so Next expects assets at `/basepath/_next/foo` but Cloudfront is serving them at `_next/foo`.

khalil-omer:
Fixes sst/sst#4935 

I just copied the approach from https://github.com/sst/sst/pull/3613 and translated it to Ion's Nextjs component. I chose not to use the `loadRoutesManifest()` to get the `basePath` because that meant bringing `Output<>` into the mix.

### DIFF
--- a/pkg/platform/src/components/aws/nextjs.ts
+++ b/pkg/platform/src/components/aws/nextjs.ts
@@ -582,6 +582,17 @@ export class Nextjs extends Component implements Link.Linkable {
             bundle: ".open-next/dynamodb-provider",
           };
         }
+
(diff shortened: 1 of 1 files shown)


khalil-omer:
No longer needed

PR #31, opened 2024-03-14
Automatically add install dir to path for fish shell

roguesherlock:
hey I added a small bit to add install dir to path for fish shell. let me know if I should anything here.
It's basically a copy of zsh config but uses a different command ([fish_add_path](https://fishshell.com/docs/current/cmds/fish_add_path.html)) to add to path for fish shell


thdxr:
thanks for taking the bait!

PR #34, opened 2024-03-17
api-gateway: add JWT auth

arjunyel:
First time with Pulumi so would like a close review, also how do I best test this? Thank you for SST friends!


fwang:
Thanks @arjunyel!

Made some changes to the args format. Here's an example - https://ion.sst.dev/docs/component/aws/apigatewayv2/#auth-jwt

fwang:
Released in 0.0.191

arjunyel:
@fwang hey boss, sorry about all the changes you had to make to my code! One thing is I had it my way originally so the auth config would be a discriminated union, meaning typescript would only let you have either jwt or iam, not both

fwang:
yeah, SST v2 used discriminated types. In Ion we decided to move away from it b/c Intellisense suggests fields from all unioned types until the discriminated field is typed out. And that caused some confusion for folks that are new to a component.

PR #37, opened 2024-03-18
feat: Allow transforming the distribution in ssr-site

cmanou:
- Allows users to transform the distribution this is useful if you want to add additional origins such as a reverse proxy to your API if its external, so you can skip CORs requests


cmanou:
Other alternative is to allow a transform for the whole CdnArgs let me know if that is preferred

cmanou:
@fwang Any thoughts on this happy to explore other solutions

cmanou:
Realized I can work around this by just putting a `Router` infront of the nextjs site so closing this for now

PR #40, opened 2024-03-19
Add newer OpenAI Embedding Models

roguesherlock:
Hey I added two newer embedding models from OpenAI, namely `text-embedding-3-small` and `text-embedding-3-large`. These are significantly cheaper and faster. You can find details about them on their announcement post [here](https://openai.com/blog/new-embedding-models-and-api-updates). Let me know if I should add anything else!


fwang:
The Postgres table uses the `hnsw` index, which supports vectors up to 2,000 dimensions - https://github.com/pgvector/pgvector?tab=readme-ov-file#what-if-i-want-to-index-vectors-with-more-than-2000-dimensions

The `text-embedding-3-large` model has a 3072 dimension. It's suggested to do a dimensionality reduction. But i'm not too familiar w/ it.

Let me add `text-embedding-3-small` for now.

fwang:
Released in 0.0.189

roguesherlock:
> The Postgres table uses the `hnsw` index, which supports vectors up to 2,000 dimensions - https://github.com/pgvector/pgvector?tab=readme-ov-file#what-if-i-want-to-index-vectors-with-more-than-2000-dimensions
> 
> The `text-embedding-3-large` model has a 3072 dimension. It's suggested to do a dimensionality reduction. But i'm not too familiar w/ it.
> 
> Let me add `text-embedding-3-small` for now.

sorry I missed that. Thank you for looking into it!

OpenAI has native support for shortening embeddings so technically we can configure the dimension size for `text-embedding-3-large` model to the max size available for pgvector. I'll look into it.

- https://openai.com/blog/new-embedding-models-and-api-updates
- https://platform.openai.com/docs/guides/embeddings/use-cases

fwang:
@roguesherlock Just added support for `text-embedding-3-large` with reduced dimension of 2000.

commit: https://github.com/sst/ion/commit/327870d96fb98655f419e4e8feb31bbc6350331f
doc: https://ion.sst.dev/docs/component/aws/vector/#model

roguesherlock:
> @roguesherlock Just added support for `text-embedding-3-large` with reduced dimension of 2000.
> 
> commit: [327870d](https://github.com/sst/ion/commit/327870d96fb98655f419e4e8feb31bbc6350331f) doc: https://ion.sst.dev/docs/component/aws/vector/#model

Awesome that was quick!

PR #47, opened 2024-03-22
Fix a typo in sns docs

roguesherlock:
Hey I believe there's a typo in sns docs since the SNSClient should be imported from `@aws-sdk/client-sns` and rather than `@aws-sdk/client-sqs`.


fwang:
Thanks @roguesherlock!

The docs are actually autogenerated from the tsdoc. Fixed in the code https://github.com/sst/ion/commit/8baf8f3f8d27ba4b571ed86e3090ba74a530a87e

PR #49, opened 2024-03-23
Follow XDG base dir spec in install script

arockett:
sst/sst#4666 

---

- [x] **Implement:** modify install script to use `$XDG_BIN_HOME` as install dir and default to `$HOME/.local/bin` if `$XDG_BIN_HOME` doesn't exist
- [x] **Implement:** modify `sst upgrade` command to use the same logic for determining the install dir
- [x] **Test:** run install script with
  - [x] $XDG_BIN_HOME set and already in $PATH
  - [x] $XDG_BIN_HOME set and not in $PATH
  - [x] $XDG_BIN_HOME not set
- [x] **Test:** verify `sst upgrade` works as expected
  - [x] build `sst` executable and manually install
  - [x] run `sst upgrade` and verify it uses the $XDG_BIN_HOME dir

### DIFF
--- a/install
+++ b/install
@@ -35,7 +35,7 @@ case "$filename" in
     ;;
 esac
 
-INSTALL_DIR=$HOME/.sst/bin
+INSTALL_DIR=${XDG_BIN_HOME:-$HOME/.local/bin}
 mkdir -p "$INSTALL_DIR"
 
 if [ -z "$requested_version" ]; then
--- a/pkg/global/upgrade.go
+++ b/pkg/global/upgrade.go
@@ -75,16 +75,11 @@ func Upgrade(existingVersion string, nextVersion string) (string, error) {
 		return "", fmt.Errorf("unexpected HTTP status when downloading release: %s", resp.Status)
 	}
 
-	homeDir, err := os.UserHomeDir()
+	tmpReleaseDir, err := os.MkdirTemp("", "sst")
 	if err != nil {
 		return "", err
 	}
-
-	sstBinPath := filepath.Join(homeDir, ".sst", "bin")
-	os.RemoveAll(sstBinPath)
-	if err := os.MkdirAll(sstBinPath, os.ModePerm); err != nil {
-		return "", err
-	}
+	defer os.RemoveAll(tmpReleaseDir)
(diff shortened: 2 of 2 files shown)


thdxr:
hey we actually overhauled our install script - do you mind fixing conflicts? this is a good fix - we also need to update our `sst upgrade` command to respect it

arockett:
Ah yes, I found the place starting @ https://github.com/sst/ion/blob/dev/pkg/global/upgrade.go#L78 to align the `sst upgrade` command with this change. I'll have time next week Monday to get to this and I'll make sure to pull in the latest install script changes.

arockett:
Ok this is ready for another review.

I want to call your attention to some details which I added in the commit description about prior sst installs:

```
An existing install in the ~/.sst/bin dir is not removed during the `sst
upgrade` command nor is the $PATH modified by `sst upgrade`.

When there is an existing install of sst prior to this commit, the
safest upgrade path is to:
1. delete the ~/.sst dir
2. remove ~/.sst/bin from $PATH by editing the right shell cfg file
3. rerun the install script
```

The most unexpected situation arises when a user has
- installed a version of sst prior to this commit
- added `~/.sst/bin` to their PATH
- does not have $XDG_BIN_HOME on their $PATH, and
- runs `sst upgrade` to get latest version instead of using the install script

In that situation, running `sst upgrade` will install the latest sst bin including this change at the old `~/.sst/bin` location... still works, then next time `sst upgrade` is run it will install the latest sst to the new location which is not on the $PATH and print to the console "upgraded to version XXX" even though running `sst version` will still pick up the old one in the `~/.sst/bin` dir.

Since it's early in the life of sst/ion I wasn't sure if you consider it important to handle this worst case situation more gracefully and want to ask your thoughts before adding extra complexity. If you're ok with just recommending a "clean install" for that situation then I think this PR is ready to merge.

If you think the worst case upgrade scenario needs a better solution, I could do something easy in the upgrade command like
- delete `~/.sst` dir if it exists
- check that install bin dir is on the $PATH and if it isn't print a message to the console

Thoughts?

thdxr:

PR #57, opened 2024-03-24
POC for cloudflare and open-next

conico974:
This is a POC for using cloudflare and aws together to deploy open-next
This is very messy at the moment as i just copied a bunch of stuff

### What's included
A cloudflare worker is used as a router to either serve static file or to redirect to either the `imageOptimizer` function in aws or the external middleware which is also on a cloudflare worker.
The middleware cloudflare worker then execute all the necessary routing (including next.config headers, redirects and rewrites) and the middleware.
It also creates all the necessary functions on lambda (image optimization, default server and all the other server functions)
It also creates an R2 bucket for the static assets.

### Requirements
A `middleware.ts` file is necessary.
It also requires some specific stuff in `open-next.config.ts` : 
```ts
import type {OpenNextConfig} from "open-next/types/open-next.js"

const config = {
  default: {
  },
  // This middleware part is necessary to make it work
  middleware: {
    external: true, 
    override: {
      wrapper: "cloudflare",
      converter: "edge"
    }
  }, 
} satisfies OpenNextConfig

export default config
```
### What's not implemented or broken
There is no automatic upload to R2, this needs to be done manually.
Somehow every fetch to the cloudflare api is broken for me (i've probably done something wrong, i have a connect timeout for every call), so in order to test i had to break the `WorkersUrl`, i then manually enabled the url in cloudflare
Edge function are not deployed, api route (from app router) should already work if implemented in the construct
No support for wasm and assets, it should work, but we need to add this to the construct.


conico974:
Replaced by #588

PR #58, opened 2024-03-24
Implemented additional functionality and script overhaul

zohaibakber:
Update installation script to add Ion to PATH in GitHub Actions environment.
Modify curl command to show download progress


zohaibakber:
Now this same script can be used for sst upgrade with very little modification

PR #66, opened 2024-03-26
fix: make custom build command work for ssr site

thomkrupa:
Currently, custom `buildCommand` does not work - tested on Remix. If I understand this correctly, buildCommand defined by user comes from args, just like environment or link.


fwang:
Thanks @thomkrupa!

Made a minor change to check for the package manager instead of always running `npm run build`

fwang:
Released in v0.0.202

PR #75, opened 2024-03-26
Better upgrade cli command with progress and other stuff...

zohaibakber:


cmd/sst/ui/ui.go
```diff
@@ -24,8 +24,9 @@ const (
 )
 
 const (
-	IconX     = "×"
+	IconX     = "x"
```
jayair: lol we spent too much time picking this `x`, how come you wanna change it?

zohaibakber: @jayair it was doing naughty stuff in my terminal, sorry for hurting your feelings :p 
Also why tf are you people spending time on a cross icon decision?

jayair: Oh really? Can I see a screenshot? I'm curious what it's doing.

zohaibakber: image

zohaibakber: maybe its my term sometimes it works sometimes it meses it up
but whatever I will revert it

pkg/global/upgrade.go
```diff
@@ -100,8 +116,21 @@ func Upgrade(version string) (string, error) {
 	return version, nil
 }
 
-func untar(reader io.Reader, target string) error {
-	tarReader := tar.NewReader(reader)
+func untar(reader io.Reader, target string, totalSize int64) error {
+	p := mpb.New(mpb.WithWidth(64))
+	bar := p.AddBar(totalSize,
```
thdxr: we do not want this function printing anything out to the terminal - everything in `pkg` should be agnostic of where it's being called from

to support a progress thing like this the `Upgrade` command would have to take a callback or something

pkg/project/create.go
```diff
@@ -79,7 +80,11 @@ func Create(templateName string, home string) error {
 				return err
 			}
 			slog.Info("patching", "file", patchStep.File)
-			data, err := os.ReadFile(patchStep.File)
+			b, err := os.ReadFile(patchStep.File)
+			if err != nil {
+				return err
+			}
+			data, err := hujson.Standardize(b)
```
thdxr: can you send this in another PR

zohaibakber: @thdxr https://github.com/sst/ion/pull/88

pkg/global/upgrade.go
zohaibakber: @thdxr I guess https://github.com/sst/ion/commit/9ae9d3d1c972cdd4318b9f8d54e2c15814f4edcf after this pr can closed?

PR #128, opened 2024-03-28
providers always loaded even when not set as home

qw-in:
My attempt to fix the issue I was having in sst/sst#4874. Literally the first go I've written so probably rough & broken 😅

What I'm trying to change is that currently only the "home" provider gets initialized:

https://github.com/sst/ion/blob/917a55c84931b88e076204f2424dbfe4e0128db1/pkg/project/project.go#L202-L212

And only "home" provider environment variables are passed to pulumi:

https://github.com/sst/ion/blob/917a55c84931b88e076204f2424dbfe4e0128db1/pkg/project/stack.go#L137

https://github.com/sst/ion/blob/917a55c84931b88e076204f2424dbfe4e0128db1/pkg/project/stack.go#L485

At which point (in my case) the cloudflare workerscript gets all mad that "accountId" or whatever is unset:

https://www.pulumi.com/registry/packages/cloudflare/api-docs/workerscript/

My idea here is to:
1. `Init()` every provider
2. Provide envvars from all providers (not just home)

I'm not sure if this fits the underlying pulumi model well, an alternative could be to just expose a global "env" config where I could set cloudflare bits etc. Or I'm dumb and it's not an issue in the first place

### DIFF
--- a/cmd/sst/main.go
+++ b/cmd/sst/main.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	flag "github.com/spf13/pflag"
 	"io"
 	"log/slog"
 	"os"
@@ -15,6 +14,8 @@ import (
 	"strings"
 	"syscall"
 	"time"
+
+	flag "github.com/spf13/pflag"
 
 	"github.com/briandowns/spinner"
 	"github.com/fatih/color"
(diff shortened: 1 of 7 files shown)


qw-in:
> ... unfortunately that PR probably is looking at the wrong approach

Via sst/sst#4874

PR #155, opened 2024-04-03
fix: Handle linux distros that report arm64 architecture as aarch64

mhodgson:
Currently the install script fails silently on Linux distros that report arm64 architecture as "aarch64". This fixes the issue.

### DIFF
--- a/install
+++ b/install
@@ -14,6 +14,10 @@ if [[ "$os" == "darwin" ]]; then
 fi
 arch=$(uname -m)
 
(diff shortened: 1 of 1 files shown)


ollieleach:
+1 - Fails on Amazon Linux arm64 ec2s

mhodgson:
@jayair any chance of getting this merged? Pretty easy fix and it's annoying to have to work around this in CI and on dev containers on ARM Macs.

jayair:
Let me ping Dax

PR #183, opened 2024-04-05
feat: IoTAuthorizer component

sammyjoyce:
An IoT custom Authorizer component, similar to what is used in sst/console


sammyjoyce:
Closing this as the new [Realtime](https://ion.sst.dev/docs/component/aws/realtime/) component makes it redundant

PR #191, opened 2024-04-06
fix: callback url replace regex

antonmoller:
I've been trying to set up github auth, but haven't been able to get it to work without setting the below in the githubadapter.
`params: {
                redirect_uri: 'https://......lambda-url.eu-north-1.on.aws/github/callback',
            },`
            
This PR should fix the issue with the regex so it now matches from `/authorize/` to EOL. Checked [sst/sst](https://github.com/sst/sst/blob/d67d9c99b1882dddb1219a482d4956d66a47c4db/packages/sst/src/node/future/auth/adapter/oauth.ts#L46) so my guess is that this is what should happen


TheKnightCoder:
This will fix sst/sst#4743

PR #229, opened 2024-04-12
Fix disabling lambda url cors error

qw-in:
Previously, disabling `url.cors` after it had been enabled would have no effect or throw an error, this attempts to fix that.

Reproduce:

1. First deploy
```ts
new sst.aws.Function("Cors", {
  handler: "cors.handler",
  url: {
    cors: true,
  },
});
```

```shell
sst deploy
sst refresh # potentially unneeded
```

2. Second deploy
```ts
new sst.aws.Function("Cors", {
  handler: "cors.handler",
  url: {
    cors: false,
  },
});
```

```shell
sst deploy
×  Failed
   Cors sst:aws:Function → Cors aws:lambda:FunctionUrl
   updating urn:pulumi:...::sst:aws:Function$aws:lambda/functionUrl:FunctionUrl::Cors: 1 error occurred:
   
   InvalidParameterValueException: You can't leave AllowOrigins as empty when Cors is enabled.
```

After this change, cors should be disabled successfully


fwang:
Thanks @qw-in!

Will be in the next release.

PR #268, opened 2024-04-18
feat: Add environment option to imageOptimization nextjs(aws)

### LINKED ISSUE #4796: feat(nextjs): Ability to pass environment variables to imageOptimization or transform the function
Would be great if either a transform or environment variables were exposed for the imageOptimization function so that the new `OPENNEXT_STATIC_ETAG` env var can be set. 

https://open-next.js.org/inner_workings/components/image_optimization 

Happy to help implement if there is a decision on best way to expose it.

cmanou:
Closes https://github.com/sst/sst/issues/4796

Alternative is to add a transform, happy to close this and do that if that would be more desired @fwang


fwang:
@cmanou we just added an `imageOptimization.staticEtag` flag — https://github.com/sst/ion/pull/272

Will be in v0.0.323

PR #266, opened 2024-04-18
feat: expose VPC config in ssr-site, include permissions when building Next.js plan

kane50613:
Allowing custom vpc configuration sst/sst#4752 

---

To access the function args before this PR merged, use the `transform` function as below:

```js
new sst.aws.Nextjs("App", {
  transform: {
    server(fn) {
      fn.vpc = {
        subnets: ["..."],
        securityGroups: ["..."],
      };

      fn.permissions = [
        // @ts-expect-error
        ...fn.permissions,
        {
          actions: ["..."],
          resources: ["..."],
        },
      ];
    },
  },
});
```

PR #270, opened 2024-04-18
feat: add vpc and related iam perms to Nextjs

danielbush:
Add `vpc` configuration option to `sst.aws.Nextjs` so that the server code in next.js can access resources in the vpc, see sst/sst#4752 .

eg

```
    new sst.aws.Nextjs(NEXTJS_COMPONENT_NAME, {
      domain,
      environment,
      // vpc requires experimental version
      vpc: {
        securityGroups: ["sg-123455678"],
        subnets: ["subnet-12345678", "subnet-87654321", "subnet-23456789"],
      },
    });
```

I've tested this by building `sst` binary (using `goreleaser` etc) on this branch and deploying it to our dev environment.  The next.js server code needs to hit elasticache through our vpc.  I've checked in AWS to verify the VPC for the "default" lambda running next js and its execution role were updated with the permission.

There's another P/R from @kane50613  sst/ion#266 .  We were both wanting to get this to happen.

### DIFF
--- a/pkg/platform/src/components/aws/nextjs.ts
+++ b/pkg/platform/src/components/aws/nextjs.ts
@@ -336,6 +336,20 @@ export interface NextjsArgs extends SsrSiteArgs {
      */
     memory?: Size;
   };
+  /**
+   * Configure the subnet and security groups for the default Lambda running the Next.js app.
+   *
+   * @example
+   * ```js
+   * {
+   *   vpc: {
+   *     subnetIds: ["subnet-12345678{f", "subnet-98765432"],
+   *     securityGroupIds: ["sg-12345678"]
+   *   }
+   * }
+   * ```
+   */
+  vpc?: SsrSiteArgs["vpc"];
 }
 
 /**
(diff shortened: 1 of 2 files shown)


danielbush:
@thdxr @fwang @jayair is there anything we could do to help get this change into `sst`?

kane50613:
gonna have conflicts if we sst/ion#266 both have different docs right

danielbush:
> gonna have conflicts if we sst/ion#266 both have different docs right

I'm happy for the maintainers to pick one (assuming they're happy for this feature to be added) so long as they 

- allow us to set subnets, security group to access a vpc
- set the required IAM permission in order to do the above

I haven't added anything to add additional ad hoc IAM permissions.  That would be a nice additional fieature.

This PR is just adding `vpc` as a config and then if it's set, adds the IAM needed to connect to the VPC.  (Not sure if I can reduce those a bit more).

jayair:
I'll ping Frank

fwang:
Thanks @danielbush!

In v0.0.324, u can set vpc like this — https://ion.sst.dev/docs/component/aws/nextjs/#vpc

PR #274, opened 2024-04-19
Fix wrong type for Function layers

JanStevens:
When working with lambda layers the `arn` property will be an `Output` type so we must allow an `Output` array or an array of `Output`


jayair:
Is there an issue for this? If not then open one? So I can assign it to Frank to take a look.

fwang:
Thanks @JanStevens!

PR #325, opened 2024-04-28
fix: change arch to arm64 if the machine is aarch64

### LINKED ISSUE #380: Install script fails on machines with aarch architecture
There is not binary in the releases for `aarch64` so the install scripts fails to download a zip. `aarch64` is the same as `arm64` so the script just needs to change `arch` to `arm64` if it equals `aarch64`.

image

### LINKED ISSUE #4656: Install script fails on machines with aarch architecture
There is not binary in the releases for `aarch64` so the install scripts fails to download a zip. `aarch64` is the same as `arm64` so the script just needs to change `arch` to `arm64` if it equals `aarch64`.

image

lukeshay:
Fixes sst/sst#4656 

image

### DIFF
--- a/install
+++ b/install
@@ -14,11 +14,14 @@ if [[ "$os" == "darwin" ]]; then
 fi
 arch=$(uname -m)
 
+if [[ "$arch" == "aarch64" ]]; then
+    arch="arm64"
+fi
+
 filename="$APP-$os-$arch.tar.gz"
 
 case "$filename" in
     *"-linux-"*)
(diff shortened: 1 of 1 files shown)


jayair:
Is there a related issue for this?

So we can assign that and have this reviewed.

lukeshay:
> Is there a related issue for this?
> 
> 
> 
> So we can assign that and have this reviewed.

No but I can create one

qw-in:
Looks similar to sst/ion#155

thdxr:
we merged the other one!

PR #343, opened 2024-04-30
docs(aws/static-site): minor typing issue

NamesMT:
Should not be HTML entity in Markdown `code` element

image


jayair:
Yeah this was breaking the docs we were generating. I need to figure out a fix for that.

PR #344, opened 2024-04-30
Email Fix: Throw Error when DNS is set but not a Domain

san4d:
This change adjusts the conditional checks for the AWS Email construct to match the error message:
> The "dns" property is only valid when "sender" is a domain

This came up while I was attempting to verify a domain:
```javascript
export const emailDomain = new sst.aws.Email('Domain', {
    sender: 'my.sub.domain',
    dns: sst.aws.dns({ zone: rootZone.id }),
})
```


fwang:
Thanks @san4d! Nice catch 👍

PR #345, opened 2024-04-30
Email Enhancement: Expose DMARC Policy on Email Component

san4d:
In the process of trying the Email component, I ran into an issue while verifying a domain because I already had a DMARC record on that domain. In order to migrate over to the SST component, I need to be able to set my own policy.

This change adds an optional input to `EmailArgs` that sets the provided policy, keeping the existing default of `v=DMARC1; p=none;`.


fwang:
Thanks @san4d

PR #369, opened 2024-05-04
fix(aws/https-redirect) default viewer certificate TLS to version 1.2

sethcarlton:
Sets the TLS version to 1.2 (recommended) for the https redirect distribution. This matches the default for the base cdn distribution as well. 

Previously was falling back to TLS 1.0 as the default

### DIFF
--- a/pkg/platform/src/components/aws/https-redirect.ts
+++ b/pkg/platform/src/components/aws/https-redirect.ts
@@ -88,6 +88,7 @@ export class HttpsRedirect extends Component {
         viewerCertificate: {
           acmCertificateArn: certificate.arn,
           sslSupportMethod: "sni-only",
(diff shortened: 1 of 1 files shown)


sethcarlton:
After looking into this further, the ssl protocol errors I'm seeing may be caused by the strictness of TLS 1.2 (2021) - I am only seeing the issue on certain public networks (like coffee shops). Still looking into it to determine what the root cause is. 

Modifying the minimum protocol for the redirect may not actually be the best approach. If it's only redirecting maybe it is better to optimize for compatibility? Or maybe use the oldest version of TLS 1.2 (2018)

jayair:
Yeah I think compatibility makes the most sense here.

sethcarlton:
Agreed. Closing this PR

PR #384, opened 2024-05-07
installer: add case for ash shell on alpine linux

### LINKED ISSUE #302: CLI installer does not support Alpine linux (ash shell)
Screenshot 2024-04-24 at 15 04 58

Currently trying to install the CLI on Alpine linux (in a CI), and seeing the error above. 

Looks like the installer is not checking for the `ash` shell case, which doesn't seem to follow the same convention as these shells:
https://github.com/sst/ion/blob/dev/install#L114-L125

### LINKED ISSUE #4719: CLI installer does not support Alpine linux (ash shell)
Screenshot 2024-04-24 at 15 04 58

Currently trying to install the CLI on Alpine linux (in a CI), and seeing the error above. 

Looks like the installer is not checking for the `ash` shell case, which doesn't seem to follow the same convention as these shells:
https://github.com/sst/ion/blob/dev/install#L114-L125

eschaefer:
Should resolve sst/sst#4719 

Since alpine might not come with an `.ashrc` or `.profile` by default, depending on the distro, it would still be up to the user to `touch ~/.ashrc && touch ~/.profile` if running this in a fresh CI environment.


eschaefer:
Will re-open without the messy git history...

PR #389, opened 2024-05-08
fix cloudflare dev attempting to run aws lambda

qw-in:
Cloudflare dev did not previously check the `warp.Runtime` for file events. So it would try to run aws lambda's and crash.

Here I've just added the same check as can be seen above for the `completeEvents` and my crashing is gone


thdxr:
you're crushing it with these PRs ty

PR #403, opened 2024-05-12
Add sst install command to steps

jericho2597:
Not sure if I missed a step elsewhere but I needed to run sst install after sst init, in order for a new Next.js app to successfully launch in dev mode, following the steps in this example.

### DIFF
--- a/www/src/content/docs/docs/start/aws/nextjs.mdx
+++ b/www/src/content/docs/docs/start/aws/nextjs.mdx
@@ -35,6 +35,16 @@ sst init
 
 This'll detect that you are in a Next.js project and create a `sst.config.ts` file in the root.
 
+#### Install Providers
+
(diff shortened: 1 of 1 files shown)


zohaibakber (commented):
This is redundant as providers are automatically installed with sst init

jericho2597:
I though that should be the case but I had to do the following to get my Next.js site to deploy:

- install SST (macos)
- sst upgrade
- npx create-next-app@latest
- cd into next-app
- sst init
- sst deploy (failed deployment with an error relating to mqtt provider)
- sst install
- sst deploy (sucessfully deployed stack)

It's possible at some point I mixed something up though

jayair:
Hmm yeah this should be automatic

jericho2597:
Will close for now, it's possible I made a mistake in the setup or unique problem on my machine

PR #421, opened 2024-05-15
aws-dart-api example

cincenfrank:
Adds a new example for deploying an AWS API with ION using Dart.

Deployed here: https://2zne28kl7c.execute-api.us-east-1.amazonaws.com/


jayair:
Very cool, I'll add some docs and list it here: https://ion.sst.dev/docs/examples/

PR #423, opened 2024-05-16
Transform<T> functions explicitly return undefined 

ravenscar:
addresses https://github.com/sst/sst/issues/4624


tallowen:
cc @fwang - since it looks like you intentionally introduced the behavior change in e7f039a58224b9c6831bfba0812fddc63e65a8fa. 

Having landed had this typecheck landed would have saved me a lot of time debugging!

jayair:
I'll ask Frank to review.

fwang:
Thanks @ravenscar @tallowen! The fix will be in the next release.

PR #425, opened 2024-05-16
fix bug to resolve case where docker context is absolute path

ravenscar:
addresses issue: https://github.com/sst/sst/issues/4623


jayair:
Hmm was this updated on our end?

PR #427, opened 2024-05-16
fix: remix action/loaders with redirect timeout

### LINKED ISSUE #395: Remix action/loaders with `redirect` return `504` error from cloudfront due to lambda timeout
using SST V: `0.0.357`

It seems Dax broke `redirect` in Remix with the latest streaming support addition. 
The following action
```ts
export async function action() {
  await db.increment()
  return redirect('/counter')
}
```
will timeout, but if you change the redirect to something like `json(some data)` it works.

## Reproduction
Deploy a new/existing project with sst, and trigger an action with a `redirect` response.
Here's a [template](https://github.com/kiliman/remix-vite-template) I used to reproduce the issue after running into it within my project

### LINKED ISSUE #4646: Remix action/loaders with `redirect` return `504` error from cloudfront due to lambda timeout
using SST V: `0.0.357`

It seems Dax broke `redirect` in Remix with the latest streaming support addition. 
The following action
```ts
export async function action() {
  await db.increment()
  return redirect('/counter')
}
```
will timeout, but if you change the redirect to something like `json(some data)` it works.

## Reproduction
Deploy a new/existing project with sst, and trigger an action with a `redirect` response.
Here's a [template](https://github.com/kiliman/remix-vite-template) I used to reproduce the issue after running into it within my project

sammyjoyce:
fixes sst/sst#4646

this issue is due to the [react-router redirect function](https://github.com/remix-run/react-router/blob/9d42655bc6eedd9a09d0c82c48406c8d9268260f/packages/router/utils.ts#L1542) that remix uses not including a body (as it's just a redirect, so is only setting headers)

streamified response with empty body ignores other response attributes, so we write a single whitespace

### DIFF
--- a/pkg/platform/functions/remix-server/regional-server.mjs
+++ b/pkg/platform/functions/remix-server/regional-server.mjs
@@ -123,8 +123,8 @@ const createApigHandler = (build) => {
       await streamToNodeStream(response.body.getReader(), responseStream);
     } else {
       writer.write(" ");
-      writer.end();
     }
+      writer.end();
   });
 };
 
--- a/pkg/platform/functions/remix-server/regional-server.mjs
+++ b/pkg/platform/functions/remix-server/regional-server.mjs
@@ -113,13 +113,16 @@ const createApigHandler = (build) => {
         "Transfer-Encoding": "chunked",
       },
     };
-    if (response.body) {
-      const reader = response.body;
-      const writer = awslambda.HttpResponseStream.from(
+
+    const writer = awslambda.HttpResponseStream.from(
         responseStream,
         httpResponseMetadata,
-      );
-      await streamToNodeStream(reader.getReader(), responseStream);
+    );
+
+    if (response.body) {
+      await streamToNodeStream(response.body.getReader(), responseStream);
+    } else {
+      writer.write(" ");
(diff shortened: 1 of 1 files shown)


pkg/platform/functions/remix-server/regional-server.mjs
```diff
@@ -113,13 +113,16 @@ const createApigHandler = (build) => {
...
-      );
-      await streamToNodeStream(reader.getReader(), responseStream);
+    );
+
+    if (response.body) {
+      await streamToNodeStream(response.body.getReader(), responseStream);
+    } else {
+      writer.write(" ");
       writer.end();
     }
```
blntrsz: Shouldn't we do a `writer.end()` in all cases?

```suggestion
      writer.write(" ");
    }
      writer.end();
```

sammyjoyce: Yes, but StreamToNodeStream already has writer.end(), so both cases already do. I'd be inclined to do away with that function and inline that branch, but left it is as is, assuming @thdxr would do that if he wanted it

thdxr (approved):

thdxr:
thank you!

Issue #361, opened 2024-05-03
AWS Cognito preSignUp trigger arn: invalid prefix


urosbelov:
Hello to all, when cognito preSignUp trigger is created, deploy is terminated and there is an error that say:

> XXX is an invalid ARN: arn: invalid prefix. Examine values at 'XXX.lambdaConfig.preSignUp'.

jayair:
Nice I'll ask Frank to review.

jakubknejzlik:
@jaduplessis I've just hit this issue and after fixing it using workaround (see below) I discovered another issue, this time in pulumi:
https://github.com/pulumi/pulumi-aws/issues/678

Should invoke permission be implemented here as well? Eg. adding invoke permission for each function:
```
new aws.lambda.Permission("AllowExecutionFromCognito", {
      action: "lambda:InvokeFunction",
      function: migrateUser.name,
      principal: "cognito-idp.amazonaws.com",
      sourceArn: pool.nodes.userPool.arn,
});
```

Workaround:
```
const migrateUser = new sst.aws.Function("MigrateUser", {
      handler: "src/lambdas/migrate-user.handler",
});
 const pool = new sst.aws.CognitoUserPool(
      "RekapUserPool",
      {
        transform: {
          userPool: {
            lambdaConfig: {
              userMigration: migrateUser.arn,
            },
         }
      }
)

jaduplessis:
@jakubknejzlik I think you make a good point. I had to implement something similar myself. I've updated the PR to create the permissions after the pool and functions have been defined

PR #447, opened 2024-05-20
fix: return valid arn from createTriggers function

jaduplessis:
fixes issue [#361](https://github.com/sst/sst/issues/4673) 

createTriggers function in file .sst/platform/src/components/aws/cognito-user-pool.ts currently returns full function object.

The triggers are then directly passed to the lambdaConfig properties on line 278. 
However, this property is expecting the arn value. Fix returns this value instead.

Cognito doesn't automatically have permission to invoke the trigger functions. So new permissions policies have to be created for each of the functions after the pool has been created

### DIFF
--- a/examples/aws-cognito/index.ts
+++ b/examples/aws-cognito/index.ts
@@ -1,0 +1,1 @@
+export async function handler() {}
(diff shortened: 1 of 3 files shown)


fwang:
Thanks @jaduplessis!

PR #458, opened 2024-05-22
Fixes #401 - Apply filter to the ENVs passed to the command

### LINKED ISSUE #401: AWS CredentialsProvider issue triggered by sst shell command
When using `set shell` to run scripts/tests that use `was-sdkV3` clients e.g. `@aws-sdk/client-dynamodb` I run into this already documented error from the sdk package https://github.com/aws/aws-sdk-js-v3/issues/2549

This appears to be being caused by the `AWS_PROFILE` value being passed through to the script from `set shell`. This has been validated to be the case with:

```bash
sst shell printenv
```

I assume that in SST V2 the `AWS_PROFILE` was not passed to the sub command of `sst bind`?

There is a user land work around as referenced in the thread above:

```javascript
// https://github.com/aws/aws-sdk-js-v3/issues/2549
if (process.env.AWS_ACCESS_KEY_ID && process.env.AWS_PROFILE) {
  delete process.env.AWS_PROFILE
}
```

This resolves the issue but feels less than ideal.

alistairstead:
Filter AWS_PROFILE from the ENVS that are passed to the sub command of `sst shell` this avoids a connection error in the was sdk.


jayair:
https://github.com/sst/sst/issues/4641

PR #460, opened 2024-05-22
Link CognitoUserPoolClient client id and secret

### LINKED ISSUE #4597: Add support for linking CognitoUserPoolClients
When using Cognito, most of the APIs require UserPoolId and/or ClientId. User pools link up just fine with resources, however the [CognitoUserPoolClient](https://ion.sst.dev/docs/component/aws/cognito-user-pool-client/) does not.

One such example is the `ConfirmSignUpCommand` from `@aws-sdk/client-cognito-identity-provider`, which requires a `ClientId`. Linking just doesn't seem to work for the client - no resource is generated in the types.generated.ts file for the client; only the user pool.

Example sst config:

```ts
const userPool = new sst.aws.CognitoUserPool("UserPool", {
    usernames: ["email"],
});
const client = userPool.addClient("Client", {
  transform: {
    client: (client) => {
      client.explicitAuthFlows = [
        "ALLOW_REFRESH_TOKEN_AUTH",
        "ALLOW_USER_SRP_AUTH",
        "ALLOW_CUSTOM_AUTH",
        "ALLOW_ADMIN_USER_PASSWORD_AUTH",
      ];
    },
  },
});

const hono = new sst.aws.Function("Hono", {
    url: true,
    handler: "functions/api/index.handler",
    link: [userPool, client],
    environment: {
        COGNITO_CLIENT_ID: client.id,
    },
});
```

In the above case, the Resource types look like this:

```ts
/* tslint:disable */
/* eslint-disable */
import "sst"
declare module "sst" {
  export interface Resource {
    Hono: {
      name: string
      type: "sst.aws.Function"
      url: string
    }
    UserPool: {
      id: string
      type: "sst.aws.CognitoUserPool"
    }
  }
}

export {}
```

You might notice that in order to work around this, I had to pipe the client id into the environment of the lambda function to access through process.env rather than `Resource`.

digitaltoad:
Adds the client `id` and `clientSecret` to links.

### DIFF
--- a/pkg/platform/src/components/aws/cognito-user-pool-client.ts
+++ b/pkg/platform/src/components/aws/cognito-user-pool-client.ts
@@ -90,6 +90,16 @@ export class CognitoUserPoolClient extends Component {
       client: this.client,
     };
   }
+
+  /** @internal */
+  public getSSTLink() {
+    return {
+      properties: {
+        id: this.id,
+        clientSecret: this.client.clientSecret,
+      }
+    }
+  }
 }
 
 const __pulumiType = "sst:aws:CognitoUserPoolClient";
--- a/pkg/platform/src/components/aws/cognito-user-pool-client.ts
+++ b/pkg/platform/src/components/aws/cognito-user-pool-client.ts
@@ -1,14 +1,9 @@
-import {
-  ComponentResourceOptions,
-  Output,
-  all,
-  interpolate,
-  output,
-} from "@pulumi/pulumi";
+import { ComponentResourceOptions } from "@pulumi/pulumi";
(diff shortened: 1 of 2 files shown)


jayair:
https://github.com/sst/sst/issues/4597

digitaltoad:
@jayair Yeah looks like this PR fixes sst/sst#4597

jayair:
Yeah just linking it so we can close that after we merge this.

fwang:
Thanks @digitaltoad!

Made a minor change to the name `clientSecret` -> `secret`

Will be in the next release.

PR #482, opened 2024-05-27
WIP: add false option to functions

yyaskriloff:
fixes issue sst/sst#4584 

I added the option to make logging false. I'm very unfamiliar with Pulumi so I hope this doesn't create more of a hassle than it solves.

I allowed args.logging to be false 
then if args.logging is false then normalize logging returns early which then doesn't create a log group with makes loggingConfig false when creating function

### DIFF
--- a/pkg/platform/src/components/aws/function.ts
+++ b/pkg/platform/src/components/aws/function.ts
@@ -414,15 +414,16 @@ export interface FunctionArgs {
    * }
    * ```
    */
-  logging?:
-    | Input<{
+  logging?: Input<
+    | {
         /**
          * The duration the function logs are kept in CloudWatch.
          * @default `forever`
          */
         retention?: Input<keyof typeof RETENTION>;
-      }>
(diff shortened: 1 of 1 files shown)


fwang:
@yyaskriloff Lambda will always create a log group for each function. Can you clarify what you are trying to do and share your use case?

yyaskriloff:
My apologies I miss understood. I didn't know lambda automatically created a log group. I guess what I'm looking for is the ability to disable the cloudwatch logs by removing the access role from lambda as to not duplicate my logs if I'm using a third party for log monitoring. 
In that case this PR can be closed because the code I wrote is far from accurate.

I thought that SST was just don't give you the option not to create not knowing that's the default. 



 [Sending AWS lambda logs to Axiom](https://axiom.co/docs/apps/lambda#disable-cloudwatch-logging)

yyaskriloff:
Would adding an explicit deny statement accomplish this?

PR #497, opened 2024-05-31
Make CognitoUserPool arn public

digitaltoad:
This makes the arn for CognitoUserPool public so it can be used in things like lambda permissions.


brunocleite (approved):

fwang:
Thanks @digitaltoad!

PR #500, opened 2024-05-31
Fix: Mkdir before symlink in copyFiles in dev mode

jamesgibbons92:
symlink was failing when the file to be copied resided in a directory. as directory was not being created properly beforehand


kevintechie:
This issue is a major blocker. It prevents us from using data and config files during dev.

jayair:
Let me ping again.

PR #511, opened 2024-06-04
Create docs for continuous integration

mattrobrob:
This PR creates a page for setting up GitHub actions to deploy an SST application.

### DIFF
--- a/pkg/platform/src/components/aws/auth.ts
+++ b/pkg/platform/src/components/aws/auth.ts
@@ -32,7 +32,6 @@ export class Auth extends Component implements Link.Linkable {
       return new Function(`${name}Authenticator`, {
         ...args,
         url: true,
-        streaming: !$dev,
         environment: {
           ...args.environment,
           AUTH_PRIVATE_KEY: secret(this.key.privateKeyPemPkcs8),
--- a/sdk/js/src/auth/handler.ts
+++ b/sdk/js/src/auth/handler.ts
@@ -65,6 +65,7 @@ export function AuthHandler<
     >;
   }[keyof Providers],
 >(input: {
+  stream?: boolean;
   session?: Sessions;
   providers: Providers;
   callbacks: {
--- a/sdk/js/src/aws/auth.ts
+++ b/sdk/js/src/aws/auth.ts
@@ -12,7 +12,7 @@ export module auth {
   >(...args: Parameters<typeof AuthHandler<Providers, Sessions>>) {
     const hono = AuthHandler(...args);
     return (
-      process.env.SST_LIVE ? handle(hono) : streamHandle(hono)
+      args[0].stream ? streamHandle(hono) : handle(hono) 
     ) as Handler<any, any>;
   }
   export const sessions = createSessionBuilder;


jayair:
Nice, yeah let me review and edit it. I might not get to it this week.

iliaamiri:
@mattkinnersley
> This PR creates a page for setting up GitHub actions to deploy an SST application.

Out of curiosity, how does it authenticate on AWS without a token? AWS Account Id is not considered sensitive data. 

Is it working with cloudfront API token?

mattkinnersley:
> @mattkinnersley
> 
> > This PR creates a page for setting up GitHub actions to deploy an SST application.
> 
> 
> 
> Out of curiosity, how does it authenticate on AWS without a token? AWS Account Id is not considered sensitive data. 
> 
> 
> 
> Is it working with cloudfront API token?

@iliaamiri 
When we set up the Identity Provider in our SST config, we give it our GitHub org and repo name. We also give it a URL which in our case is the generic GitHub actions URL. These tell AWS that if anything is trying to assume a role in our account from a GitHub action running in our exact repo, then allow it. So for deploying to AWS from a GitHub action, we can use short lived tokens that are generated when the action runs. To the best of my knowledge, Cloudflare doesn't have this kind of Identity Provider/ Assume Role mechanism, so we simply provide a token to the GitHub action which should be scoped to do only what you need it to do.

iliaamiri:
@mattkinnersley 

Appreciated the thorough explanation. Makes total sense. I guess It'd be great to also add instructions for IAM Identity Center too for beginners like me. (Instructions on how to setup the AWS side basically)

xinha-sh:
@mattkinnersley Should the thumbprintLists value be stored as a secret ?
```ts
const oidc = new aws.iam.OpenIdConnectProvider("GithubOidc", {
    clientIdLists: ["sts.amazonaws.com"],
    thumbprintLists: ["aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"],
    url: `https://token.actions.githubusercontent.com`,
  });
  ```

**Comment by mattkinnersley** [comment]
> @mattkinnersley Should the thumbprintLists value be stored as a secret ?
> 
> ```ts
> 
> const oidc = new aws.iam.OpenIdConnectProvider("GithubOidc", {
> 
>     clientIdLists: ["sts.amazonaws.com"],
> 
>     thumbprintLists: ["aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"],
> 
>     url: `https://token.actions.githubusercontent.com`,
> 
>   });
> 
>   ```

@xinha-sh No you don't need to store it as a secret, the value is actually completely arbitrary and could be anything. AWS has GitHub as a recognised provider and already knows the thumbprint. Pulumi, however, haven't updated their SDK to reflect this value isn't needed so we need to provide a value just to satisfy the expected type.

**Comment by xinha-sh** [comment]
@mattkinnersley This documentation basically uses one AWS org for different stages right ? 
What changes needs to be done in order to use different aws organization / child organization [aws-sso](https://docs.sst.dev/working-with-your-team#aws-account-per-environment-and-developer) for different stages ?

File: www/src/content/docs/docs/continuous-integration.mdx [hunk 1]
```diff
@@ -0,0 +1,127 @@
...
+---
+
+## OIDC Provider and Role
+
+In your `sst.config.ts`, add the following code:
+
+```ts
+const isPersonalStage = $app.stage !== "production" && $app.stage !== "dev";
+if (!isPersonalStage) {
+  const oidc = new aws.iam.OpenIdConnectProvider("GithubOidc", {
```
xinha-sh: This throws an error when deployment to another stage is attempted.

Ex: When deploying for this first time with stage dev this works fine. When attempting deployment with stage production this errors out as this is already present. 

I checked pulumi docs if we can fetch oidc connect provider [https://www.pulumi.com/registry/packages/aws/api-docs/iam/openidconnectprovider/#look-up](https://www.pulumi.com/registry/packages/aws/api-docs/iam/openidconnectprovider/#look-up)

```ts
public static get(name: string, id: [Input<ID>](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/#ID), state?: OpenIdConnectProviderState, opts?: [CustomResourceOptions](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/#CustomResourceOptions)): OpenIdConnectProvider
```

As you can see it requires ID as second argument to be fetched. Not sure if ARN will work or thumbprint.

iliaamiri:
> This PR creates a page for setting up GitHub actions to deploy an SST application.

Hey man, I just wanted to appreciate this documentation. I just got the company's Next.js CD pipelines automated in GitHub action, and I'm so happy about it. Huge thanks to you! You rock! :partying_face:

www/src/content/docs/docs/continuous-integration.mdx
mattrobrob: If you're using the same AWS account for prod and dev, that's where you'll have issues. The providers aren't namespaced and exist at a global level in the AWS account. I only deploy this to production as I don't want my github action deploying to a dev stage, which is why I haven't run into this issue.

mattkinnersley:
> @mattkinnersley This documentation basically uses one AWS org for different stages right ? What changes needs to be done in order to use different aws organization / child organization [aws-sso](https://docs.sst.dev/working-with-your-team#aws-account-per-environment-and-developer) for different stages ?

@xinha-sh No this works with AWS SSO which is what I use. You just need to set the correct account number as a variable in your GitHub repo settings.

zammitjames (approved):
🔥

www/src/content/docs/docs/continuous-integration.mdx
xinha-sh: @mattkinnersley Yup figured that out, forgot to mention.
I'll still suggest this to be:
```ts
const oidc = new aws.iam.OpenIdConnectProvider(`GithubOidc_${$app.stage}`,{})
```
Since the documentation will be read by users who'll be most probably starting up with single AWS account and they may encounter this error

mattrobrob: SST is usually pretty good at prefixing components with the stage elsewhere- I guess that's not the case here?

iliaamiri:
## I've discovered a few possible caveats
My custom use case let me discover some caveats that people might end up experiencing:

1. If you remove any stage of your app, it can possibly remove the OIDC as well. This can cause the rest of the stages to not be able to authenticate on AWS in the runner (on GitHub actions) because there is no Identity Providers anymore.
**Possible solution**: manually create the GitHub OIDC on your IAM console and get its ARN like this:
```typescript
const oidc = await aws.iam.getOpenIdConnectProvider({
  url: 'https://token.actions.githubusercontent.com',
});

// oidc.arn

```

2. The name of the GitHub role is too generic. The role being generic limits you with how many different projects you want to use this for. 
**Possible solution**: name the role as `${$app.stage}-${$app.name}-GithubRole`.

It would be great if we could tell SST not to remove these things specifically.

### Update
I ended up manually creating the github roles on AWS. This was a requirement for my case because in the long run, removing a stage would remove the role. And can potentially cause outage in the pipeline automation. 

That's not to say this isn't helpful. I couldn't do this without these docs. 🙏

mattkinnersley:
> ## I've discovered a few possible caveats
> 
> My custom use case let me discover some caveats that people might end up experiencing:
> 
> 
> 
> 1. If you remove any stage of your app, it can possibly remove the OIDC as well. This can cause the rest of the stages to not be able to authenticate on AWS in the runner (on GitHub actions) because there is no Identity Providers anymore.
> 
> **Possible solution**: manually create the GitHub OIDC on your IAM console and get its ARN like this:
> 
> ```typescript
> 
> const oidc = await aws.iam.getOpenIdConnectProvider({
> 
>   url: 'https://token.actions.githubusercontent.com',
> 
> });
> 
> 
> 
> // oidc.arn
> 
> 
> 
> ```
> 
> 
> 
> 2. The name of the GitHub role is too generic. The role being generic limits you with how many different projects you want to use this for. 
> 
> **Possible solution**: name the role as `${$app.stage}-${$app.name}-GithubRole`.
> 
> 
> 
> It would be great if we could tell SST not to remove these things specifically.
> 
> 
> 
> ### Update
> 
> I ended up manually creating the github roles on AWS. This was a requirement for my case because in the long run, removing a stage would remove the role. And can potentially cause outage in the pipeline automation. 
> 
> 
> 
> That's not to say this isn't helpful. I couldn't do this without these docs. 🙏 

Thanks @iliaamiri those are good points. I think these docs will likely be improved by pointing people towards using the SST console with all the work they're doing on autodeploy

mattkinnersley:
Closing at Console now supports CI

PR #524, opened 2024-06-07
Fix aws-monorepo example script types and core index export

rmarscher:
I started building yesterday off the latest aws-monorepo example and noticed the scripts folder sst-env.d.ts was still pointing to the old type file.

I also noticed I couldn't import from a packages/core/src/index.ts file without adding an entry to the exports in package.json. 

Thanks!

### DIFF
--- a/examples/aws-monorepo/packages/core/package.json
+++ b/examples/aws-monorepo/packages/core/package.json
@@ -2,6 +2,7 @@
   "name": "@aws-monorepo/core",
   "version": "0.0.0",
   "exports": {
+    ".": "./src/index.ts",
     "./*": [
       "./src/*/index.ts",
       "./src/*.ts"
--- a/examples/aws-monorepo/packages/scripts/sst-env.d.ts
+++ b/examples/aws-monorepo/packages/scripts/sst-env.d.ts
@@ -1,1 +1,3 @@
-/// <reference path="../../.sst/types.generated.ts" />
+/* tslint:disable */
+/* eslint-disable */
+/// <reference path="../../sst-env.d.ts" />


rmarscher:
I'm finding a related issue with this workspace / tsconfig / package.json exports configuration:

If I create `packages/core/src/utils.ts` and then have an import from `'@aws-monorepo/core/utils'`. It's fine in my vscode type checker but when running with sst dev, the bundler says it can't find that module.

If I move it to `packages/core/src/utils/index.ts`, it works. If I find a fix before this PR closes, I'll update.

rmarscher:
```json
  "exports": {
    ".": "./src/index.ts",
    "./*": [
      "./src/*.ts"
    ]
  }
```
works fairly well. You have to import `@aws-monorepo/core/example/index` or move example/index.ts to example.ts. 

Still searching for a solution that supports all use cases. Understand this isn't really an SST Ion issue - although maybe there is something happening in the esbuild config that causes the module resolution to be different.

jayair:
Yeah let me know if you find something for the index export issue.

rmarscher:
I have tried to find documentation on the array option for subpath exports in the esbuild and node.js docs but didn't see a solution. It seems like there's a general push towards using full import paths with extensions rather than rely on index resolution or automatic file extensions. In my personal project, I've been using the config from my comment and that works well for me. 

I can set it to 
```json
  "exports": {
    ".": "./src/index.ts",
    "./*": [
      "./src/*.ts",
      "./src/*/index.ts"
    ]
  }
```
and it will continue to work for any non-index subpath entries but still won't work for `@aws-monorepo/core/example`. The first array entry (`"./src/*.ts"`) seems to always be used without falling back to the second entry (`"./src/*/index.ts"`).


Feel free to close this PR and handle it all another way. Thanks!

rmarscher:
Ah... here - https://github.com/evanw/esbuild/issues/3536 and https://github.com/evanw/esbuild/issues/2974

rmarscher:
I'm going to go ahead and close this. There's a disconnect between the way Typescript handles the exports field and the way esbuild handles it.

PR #541, opened 2024-06-11
fix: make SsrSite server cache policy maxTtl 1 year

luke-rucker:
Makes SsrSite's server cache policy `maxTtl` 1 year in seconds. Based on the 365 value, I think this was the intent, just the wrong unit.


jayair:
Thanks!

PR #549, opened 2024-06-13
fix(node): support npm versions 10 and above

### LINKED ISSUE #578: AWS Lambda Deployment Fails on M1 Mac with npm v10
## Environment
M1 Mac
npm version v10

* sst.config.ts
```ts
new sst.aws.Remix("CuratorClip", {
  transform: {
    server: {
      nodejs: {
        install: ["@node-rs/argon2", "@node-rs/bcrypt"],
      },
      architecture: "arm64",
    },
   },
});
```

## Problem
I'm using the `@node-rs/*` module.

Deployments succeed, but the application is not accessible. This issue occurs because the node_modules directory uploaded to AWS Lambda is built for Darwin architecture. Starting from npm version 9, the arch and platform options have been changed to cpu and os.

* `--platform` to `--os`
* `--arch ` to `--cpu`

https://docs.npmjs.com/cli/v9/commands/npm-install#cpu

### LINKED ISSUE #4509: AWS Lambda Deployment Fails on M1 Mac with npm v10
## Environment
M1 Mac
npm version v10

* sst.config.ts
```ts
new sst.aws.Remix("CuratorClip", {
  transform: {
    server: {
      nodejs: {
        install: ["@node-rs/argon2", "@node-rs/bcrypt"],
      },
      architecture: "arm64",
    },
   },
});
```

## Problem
I'm using the `@node-rs/*` module.

Deployments succeed, but the application is not accessible. This issue occurs because the node_modules directory uploaded to AWS Lambda is built for Darwin architecture. Starting from npm version 9, the arch and platform options have been changed to cpu and os.

* `--platform` to `--os`
* `--arch ` to `--cpu`

https://docs.npmjs.com/cli/v9/commands/npm-install#cpu

gronxb:
## Environment
M1 Mac
npm version 10

* sst.config.ts
```ts
new sst.aws.Remix("CuratorClip", {
  transform: {
    server: {
      nodejs: {
        install: ["@node-rs/argon2", "@node-rs/bcrypt"],
      },
      architecture: "arm64",
    },
   },
});
```

## Problem
I'm using the `@node-rs/*` module.

Deployments succeed, but the application is not accessible. This issue occurs because the node_modules directory uploaded to AWS Lambda is built for Darwin architecture. Starting from npm version 9, the arch and platform options have been changed to cpu and os.

* `--platform` to `--os`
* `--arch ` to `--cpu`

https://docs.npmjs.com/cli/v9/commands/npm-install#cpu

## Solution
To handle backward compatibility and the latest versions in npm, you can use duplicate options. It doesn't matter if there are duplicates.

Fixed sst/sst#4509


jayair:
Interesting. Yeah we'll take a look.

pkg/platform/src/runtime/node.ts
```diff
@@ -160,12 +160,17 @@ export async function build(
...
+        // support npm versions 10 and above
+        "--os=linux",
+        input.architecture === "arm64" ? "--cpu=arm64" : "--cpu=x64",
+      ];
       if (installPackages.includes("sharp")) {
-        cmd.push(
-          "--os=linux --libc=glibc",
-          input.architecture === "arm64" ? "--arch=arm64" : "--arch=x64",
-        );
+        cmd.push("--libc=glibc");
```
gronxb, line 173: I'm not sure if it's really value using the `-libc` flag just for `sharp`. I expect it to work even without that line.
Therefore, I'd like to ask if i could remove L172-174.

https://github.com/lovell/sharp/issues/2230#issuecomment-1509870904

gronxb: image
It will result in exactly the same binaries whether the flag is used or not.

```sh
$ npm install --cpu=x64 --os=linux sharp
$ npm install --cpu=x64 --os=linux --libc=glibc sharp
```

PR #569, opened 2024-06-17
Externalize all imported packages in sst.config.ts

### LINKED ISSUE #567: External libraries should not be included in eval-*.mjs output
When writing custom plugins for esbuild or including other packages in your sst.config.ts it should not bundle them into the `eval-*.mjs` file.

https://github.com/sst/ion/blob/dev/pkg/js/js.go#L39

Maybe we should set External to * or use `Packages: esbuild.PackagesExternal`?

### LINKED ISSUE #4515: External libraries should not be included in eval-*.mjs output
When writing custom plugins for esbuild or including other packages in your sst.config.ts it should not bundle them into the `eval-*.mjs` file.

https://github.com/sst/ion/blob/dev/pkg/js/js.go#L39

Maybe we should set External to * or use `Packages: esbuild.PackagesExternal`?

leon:
So that it doesn't try to inline the whole of typescript when writing custom esbuild plugins 😅

fixes sst/sst#4515


leon:
We had to roll this one back because of how some people where using NX or other mono repo setups to include reusable sst constructs from a @myorg/infra which pointed to a local monorepo package.

From what I understood this caused esbuild to think that it was an external package and maybe because it was pointing to a typescript file instead of a compiled file, it gave the error

```
Error       
|  TypeError: Unknown file extension ".ts" for /Users/user/my-app/infra/mod.ts
|      at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:160:9)
|      at defaultGetFormat (node:internal/modules/esm/get_format:203:36)
|      at defaultLoad (node:internal/modules/esm/load:143:22)
|      at ModuleLoader.load (node:internal/modules/esm/loader:396:7)
|      at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:278:45)
|      at link (node:internal/modules/esm/module_job:78:21) {
|    code: 'ERR_UNKNOWN_FILE_EXTENSION'
|  }
```

I'm not sure how we can get esbuild to only use external packages, but only if they are not local?

PR #576, opened 2024-06-19
fix(docs): incorrect CORS attribute names

DougJod:
Documentation on page:
https://ion.sst.dev/docs/component/aws/function/#url-cors
Corrected three instances where attribute name was 'allowed' instead of 'allow'
e.g. 'allowedOrigins' => 'allowOrigins'


jayair:
Thanks!

PR #588, opened 2024-06-21
Next on cloudflare

conico974:
This PR add partial support for Next on cloudflare.
For now only the middleware and the assets can be deployed on cloudflare, the rest needs to be on aws.

It requires this https://github.com/sst/open-next/pull/449 to be merged to work perfectly ( Likely in 3.0.6 )

The assets are deployed on KV and all the cache related stuff are still in aws.

One big issues i've found while testing was that it seems impossible to invalidate cache when using only worker url. To fix this the cache is disabled by default when no domains is set.

Below is an `open-next.config.ts` file that works (and is necessary) to deploy : 
```ts
import type {OpenNextConfig} from 'open-next/types/open-next.js'

const config  = {
  default: {
    override: {
      // This is not necessary, but useful to test streaming
      wrapper: 'aws-lambda-streaming',
    }
  }, 

  middleware: {
    external: true,
    override: {
      wrapper: 'cloudflare',
      converter: 'edge'
    }
  },
  imageOptimization: {
    // @ts-ignore
    loader: 'host',
    arch: 'x64'
  },
  // buildCommand: 'echo "Skipping build..."'
} satisfies OpenNextConfig

export default config
```

### DIFF
--- a/pkg/platform/functions/cf-ssr-site-router-worker/index.ts
+++ b/pkg/platform/functions/cf-ssr-site-router-worker/index.ts
@@ -2,13 +2,19 @@ declare var caches: any;
 declare var SST_ASSET_MANIFEST: Record<string, string>;
 declare var SST_ROUTES: { regex: string; origin: "assets" | "server" }[];
 
+import { Buffer } from 'node:buffer';
+
 export interface Env {
   ASSETS: any;
   SERVER: any;
+  ASSETS_PREFIX?: string;
 }
-
 export default {
   async fetch(request: Request, env: Env): Promise<Response> {
+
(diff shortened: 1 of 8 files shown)


thdxr:

conico974:
> we are moving `sst/ion -> sst/sst` and unfortunately this is going to be a messy process because of github's limitations
> 
> i'm going to close this PR for now and it can be re-opened in the new `sst/sst` repo that will be ready in the next few days if still relevant

Not sure what to do with this one. Given that now there is an official cloudflare OpenNext it makes sense that this should be at least renamed and moved to avoid confusion.

The ideal thing would be to rework the current NextJs construct and allow to have either cloudflare or cloudfront as a cdn in front (something generic would be the best so that every framework could use it) and also add support for ecs since this is supported in OpenNext itself. @thdxr what's your take on this ?

PR #591, opened 2024-06-24
fix inconsistency with AWS being top level and cloudflare being nested

versecafe:
Just fixes how `/docs/start/aws/[slug]/` were top level in the sidebar but `/docs/start/cloudflare/[slug]/` were nested under a second tier, small thing that was bothering me when looking through, should also make it cleaner as more providers get added.

## Old
image

## New
image


jayair:
Holding off on this for now since AWS is far and away the default. But I'll do this eventually.

PR #593, opened 2024-06-24
feat(aws/ssrsite): allow user singletonCachePolicy

jmeltin:
Allow a user to use specify their own singletonCachePolicy id instead of creating a new cache policy for each ssr site project.

singletonCachePolicy is added as a new argument in SsrSiteArgs
Screenshot 2024-06-24 at 9 52 03 AM

If a singletonCachePolicy is defined then it will use the defined id. Otherwise it will create a new cache policy and use the new policies id.
Screenshot 2024-06-24 at 9 52 19 AM

This has been tested locally using Nuxt. When the new Nuxt component was created it used the specified cache policy id instead of creating a new cache policy. When the Nuxt component was removed from the project it did not delete the cache policy associated with the specified id.


jayair:
Interesting, something like this is required.

We might need to review the naming.

jmeltin:
My team has hit the hard limit of 20 cache policies in the past and a requirement for us to use sst is to have a way to share policies.

I'm happy to update the name to something more fitting.

iliaamiri:
My team is having similar experiences as @jmeltin has.

eschaefer:
Is this something that would also possibly resolve https://github.com/sst/sst/issues/4667?

iliaamiri:
> Is this something that would also possibly resolve sst/sst#4667?

Yes, this will indeed fix this issue.

michaelgmcd:
Thoughts on making this the default behavior? I'd assume most people at a decent scale (my team included) need this. I know SST hasn't historically done account-level singletons, but we've ended up doing this for several things in v2 for simplicity and account limits. Log retention IAM roles being another one that comes to mind.

eschaefer:
Anything we can help with to move this along?

jmeltin:
I'm happy to rename the argument to better fit the rest of the args. @jayair Thoughts? I chose singletonCachePolicy because it most closely matched its purpose.

jayair:
We just need to review the design internally. It's a bit different from what we have so far. But we are looking into it shortly.

fwang:
Thanks @jmeltin! Renamed it to `cachePolicy` to keep it short - https://ion.sst.dev/docs/component/aws/nextjs/#cachepolicy

Will be in the next release.

PR #594, opened 2024-06-24
Allows Compatibility with Newer Vercel TF & Pulumi Provider

versecafe:
Adds compatibility for vercel terraform versions now supporting edge configs, log drains, swapping git repo without deleting existing vercel proj and losing all related logs analytics etc, lambda size configs, and more.

This is needed as currently you can't use a separate version of the pulumi provider at the same time and the `1.3.0` is missing a large number of important resources and fields on them as well as missing one of the few reasonable reasons to use vercel over AWS or CF, being edge configs


jayair:
Oh will review

PR #597, opened 2024-06-25
feat: Add ability to install a specific version with the install script

### LINKED ISSUE #572: Install a specific version of SST
👋 Hi

Is there a way to specify the SST version to be installed when using the install script `https://ion.sst.dev/install`?

I see in #77, khalil-omer has provided a GitHub action workflow which installs a specific version which in the interim I'll adopt for our project

Similarly, is there a SST version lockfile or similar that we can commit to the root of a project to ensure all developers use the same SST version?  (or at least are prompted to update when they're woefully out of date?)

Perhaps something like an `.nvmrc` for SST would be a welcome addition with regards to the above so that if a user or CI runs SST install in a project with an `.sstrc` file that specifics a version, it defaults to installing that version rather than latest.

Thanks

### LINKED ISSUE #4511: Install a specific version of SST
👋 Hi

Is there a way to specify the SST version to be installed when using the install script `https://ion.sst.dev/install`?

I see in sst/ion#77, khalil-omer has provided a GitHub action workflow which installs a specific version which in the interim I'll adopt for our project

Similarly, is there a SST version lockfile or similar that we can commit to the root of a project to ensure all developers use the same SST version?  (or at least are prompted to update when they're woefully out of date?)

Perhaps something like an `.nvmrc` for SST would be a welcome addition with regards to the above so that if a user or CI runs SST install in a project with an `.sstrc` file that specifics a version, it defaults to installing that version rather than latest.

Thanks

mhodgson:
Fixes sst/sst#4511

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,12 @@ _Note: Ion is generally available and recommended for new SST users. We are work
 curl -fsSL https://ion.sst.dev/install | bash
 ```
 
+To install a specific version:
+
+```
+curl -fsSL https://ion.sst.dev/install | VERSION=0.0.403 bash
+```
+
 To use a package manager, [check out our docs](https://ion.sst.dev/docs/reference/cli/).
 
 #### Manually
--- a/install
+++ b/install
@@ -8,6 +8,8 @@ YELLOW='\033[1;33m'
 ORANGE='\033[38;2;255;140;0m'
 NC='\033[0m' # No Color
 
+requested_version=${VERSION:-}
+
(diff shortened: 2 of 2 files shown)


jayair:
Thanks!

jayair:
For the docs, instead of the README, can you move it to here: https://github.com/sst/ion/blob/dev/cmd/sst/main.go#L175

That updates the docs site.

mhodgson:
@jayair looks like Dax pulled the trigger already. I'll let you guys sort the doc updates out.

PR #606, opened 2024-06-26
adds findRecord to cloudflare only

MrgSub:
This PR adds `findRecord` function to Cloudflare, currently they're the only provider that supports the feature.

**Use case:** 
Re-running ion would error if a record already exists, it helps if we can check to avoid this error.

### DIFF
--- a/pkg/platform/src/components/vercel/dns.ts
+++ b/pkg/platform/src/components/vercel/dns.ts
@@ -70,6 +70,7 @@ export function dns(args: DnsArgs) {
   return {
     provider: "vercel",
     createRecord,
+    findRecord
   } satisfies Dns;
 
   function useCAARecord(namePrefix: string, opts: ComponentResourceOptions) {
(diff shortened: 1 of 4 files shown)


pkg/platform/src/components/vercel/dns.ts
```diff
@@ -89,6 +90,12 @@ export function dns(args: DnsArgs) {
     return caaRecord;
   }
 
+  function findRecord() {
+    throw new Error(
+      `Currentr version of the Vercel adapter does not support finding records.`
```
ImLunaHey: typo on `Currentr` -> `Current`

jayair:
Oh interesting. Let me ask the team to take a look.

fwang:
Thanks @MrgSub!

Pulumi/Terraform provides a way to ignore existing records. You can set it through `transform`:
```ts
sst.cloudflare.dns({
  zone: "415e6f4653b6d95b775d350f32119abb",
  transform: {
    record: {
      allowOverwrite: true,
    }
  }
})
```
Does that work for you?

fwang:

PR #614, opened 2024-06-29
Allow for configuring shared Server Cache Policy ID

iliaamiri:
### Addresses this issue: https://github.com/sst/sst/issues/4667

AWS has a hard limit of 20 custom Cache Policies. My use case heavily depends on this. I've have been using [SST's CDK Server Cache Policy's configuration](https://docs.sst.dev/constructs/NextjsSite#reusing-cloudfront-cache-policies). Yet, in SST Ion, this is the _last_ piece that's blocking me from migrating. 

P.S.: I just rediscovered [this PR](https://github.com/sst/ion/pull/593) late which does the same thing.

### DIFF
--- a/pkg/platform/src/components/aws/nextjs.ts
+++ b/pkg/platform/src/components/aws/nextjs.ts
@@ -373,6 +373,18 @@ export interface NextjsArgs extends SsrSiteArgs {
    * ```
    */
   vpc?: SsrSiteArgs["vpc"];
+  /**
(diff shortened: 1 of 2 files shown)


fwang:
Thanks @iliaamiri! Fix in sst/ion#593 

https://ion.sst.dev/docs/component/aws/nextjs/#cachepolicy

Will be in the next release

PR #619, opened 2024-06-29
Update Next example

VirtuaBoza:
Adds the webpack config override to get around an issue when a dependency depends on https://www.npmjs.com/package/process, causing the middleware compilation to load the browser shim.

### DIFF
--- a/examples/aws-nextjs/next.config.mjs
+++ b/examples/aws-nextjs/next.config.mjs
@@ -1,4 +1,14 @@
+import * as path from "node:path";
+
 /** @type {import('next').NextConfig} */
-const nextConfig = {};
+const nextConfig = {
+  webpack: (config, { nextRuntime }) => {
+    if (nextRuntime === "edge") {
+      // Ensure sst uses Node's process module and not a popular shim package when compiling middleware.
+      config.resolve.alias.process = path.resolve("node:process");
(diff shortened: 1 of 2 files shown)


Wundero (approved):

jayair:
This is specifically for middleware?

VirtuaBoza:
> This is specifically for middleware?

yes

jayair:
I see. Can you give some ways I can test this? When does this happen?

VirtuaBoza:
> I see. Can you give some ways I can test this? When does this happen?

Sorry it took so long to get around to this. Here's an example of the issue: https://github.com/VirtuaBoza/sst-process-repro/blob/main/src/middleware.ts#L5-L11

Attempting to access sst Resources in Next middleware will throw "X is not linked" because an NPM package called "process" is installed in node_modules and sst is importing from "process" without specifying "node:process" due to issues with other runtimes. Next's middleware/edge webpack config is picking up the browser implementation from that package.

The NPM package I'm referencing is: https://www.npmjs.com/package/process
That package is a dependency of some popular libraries such as Cypress.

jayair:
Oh I see. Is there a downside to SST using `node:process` internally instead of `process`?

fwang:

PR #634, opened 2024-07-03
Secure Function Urls with IAM Auth

cgcompassion:
PR for issue: https://github.com/sst/sst/issues/4483

The Nextjs construct makes lambdas that are publicly available at a function url. We need the option to lock those down with IAM.

Note, here is the equivalent functionality in SST 2: https://docs.sst.dev/constructs/NextjsSite#regionalenableserverurliamauth


cgcompassion:
This can be closed now. Implemented seperately. Related: https://github.com/sst/sst/issues/4684

PR #638, opened 2024-07-04
fix: set correct assetsVersionedSubDir for remix

### LINKED ISSUE #4437: Remix: static assets are not served with long-term caching
Static assets for a remix deployment are served with the `nonVersionedFilesCacheHeader` and don't benefit from efficient caching.

Bildschirmfoto 2024-07-04 um 17 07 27

This seems to happen because the `assetsVersionedSubDir` isn't set correctly for vite builds of remix. Vite puts the static assets in `build/client/assets`.

Issue for PR sst/ion#638

klaemo:
Fixes: sst/sst#4437 

Static assets for a remix deployment are served with the `nonVersionedFilesCacheHeader` and don't benefit from efficient caching.

Bildschirmfoto 2024-07-04 um 17 07 27

This seems to happen because the `assetsVersionedSubDir` isn't set correctly for vite builds of remix. Vite puts the static assets in `build/client/assets`.

Bildschirmfoto 2024-07-04 um 17 03 14

This PR sets the directory correctly (I think) which should result in the correct headers being applied (I hope). Literally only been playing with SST for 30 minutes. No idea if that's the right place. Wouldn't be surprised if I'm showing severe skill issues here :D

### DIFF
--- a/platform/src/components/aws/remix.ts
+++ b/platform/src/components/aws/remix.ts
@@ -435,7 +435,7 @@ export class Remix extends Component implements Link.Linkable {
           const assetsPath = isUsingVite
             ? path.join("build", "client")
             : "public";
-          const assetsVersionedSubDir = isUsingVite ? undefined : "build";
+          const assetsVersionedSubDir = isUsingVite ? "assets" : "build";
 
           return {
             assetsPath,
--- a/platform/src/components/cloudflare/remix.ts
+++ b/platform/src/components/cloudflare/remix.ts
@@ -267,7 +267,7 @@ export class Remix extends Component implements Link.Linkable {
           const assetsPath = isUsingVite
             ? path.join("build", "client")
             : "public";
-          const assetsVersionedSubDir = isUsingVite ? undefined : "build";
+          const assetsVersionedSubDir = isUsingVite ? "assets" : "build";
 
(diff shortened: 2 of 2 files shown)


jayair:
Oh interesting. Is there an issue associated with this? I can assign it to the team to take a look.

klaemo:
> Is there an issue associated with this?

No, but I could create one. It would have the same content as this PR, though...

jayair:
Yeah I think do it anyway. We go through all the issues assigned to us.

platform/src/components/aws/remix.ts
```diff
@@ -435,7 +435,7 @@ export class Remix extends Component implements Link.Linkable {
           const assetsPath = isUsingVite
             ? path.join("build", "client")
```
klaemo, line 436: I think this should actually use the [buildDirectory](https://remix.run/docs/en/main/file-conventions/vite-config#builddirectory) property 🤔 

Edit: see issue https://github.com/sst/ion/issues/688

klaemo: This is handled in #736

klaemo:
imho, this PR can be closed because my new one (#736) handles this in a more generalized way

PR #648, opened 2024-07-08
add missing sst import

kleenkanteen:
the `npm install` package missed `sst`. also changed the order to have the imports first so no errors show when adding the code (confused me why I was told to add code that would show errors at first).

### DIFF
--- a/www/src/content/docs/docs/start/aws/hono.mdx
+++ b/www/src/content/docs/docs/start/aws/hono.mdx
@@ -119,7 +119,13 @@ export const handler = handle(app);
 We are directly accessing our S3 bucket with `Resource.MyBucket.name`.
 :::
 
-Add the relevant imports.
+Install the npm packages.
+
+```bash
+npm install hono sst @aws-sdk/client-s3 @aws-sdk/s3-request-presigner
+```
+
+Then add the relevant imports.
 
 ```ts title="index.ts"
 import { Resource } from "sst";
@@ -136,11 +142,6 @@ import { handle } from "hono/aws-lambda";
 const s3 = new S3Client({});
 ```
 
-And install the npm packages.
-
-```bash
-npm install hono @aws-sdk/client-s3 @aws-sdk/s3-request-presigner
-```
 
 ---
 


kleenkanteen:
Also I had an issue but this was this fix: https://github.com/sst/sst/issues/4523

Should I add that in the PR as well?

jayair:
I'm confused about your other comment. What was the fix?

kleenkanteen:
The current docs write `npm install hono @aws-sdk/client-s3 @aws-sdk/s3-request-presigner`. it forgets to add sst as well, causing an import error when it is imported 

My second comment is another issue. I can open that in another PR if you want.

kleenkanteen:
Regarding my second comment, it is a separate issue from my first comment. 

Following the guide, I got this broken code that would run the first time, but when I stopped the sst cli and reran it, it gave that error I mention in the other post. Here is the code::

```
/// <reference path="./.sst/platform/config.d.ts" />

export default $config({
    app(input) {
      return {
        name: "backend",
        removal: input?.stage === "production" ? "retain" : "remove",
        home: "aws",
      };
    },
    async run() {
    

      const hono = new sst.aws.Function("Hono", {
        url: true,
        link: [bucket],
        handler: "index.handler",
      });
  
      return {
        api: hono.url,
      };
    }
  });
 
const bucket = new sst.aws.Bucket("MyBucket", {
  public: true
});            
```

Changing it by moving the bucket declaration at the bottom inside the run() function fixed it:
```
/// <reference path="./.sst/platform/config.d.ts" />

export default $config({
    app(input) {
      return {
        name: "backend",
        removal: input?.stage === "production" ? "retain" : "remove",
        home: "aws",
      };
    },
    async run() {
    const bucket = new sst.aws.Bucket("MyBucket", {
        public: true
        });            

      const hono = new sst.aws.Function("Hono", {
        url: true,
        link: [bucket],
        handler: "index.handler",
      });
  
      return {
        api: hono.url,
      };
    }
  });
  
```

jayair:
Oh I see. Let's leave that for now.

kleenkanteen:
Why leave it out? I was stuck till I tried that since the issue was with the bucket creation

PR #652, opened 2024-07-09
Correct DynamoDB range key documentation for primary indices

ahjmorton:
Minor change to correct a copy pasted description


jayair:
Thanks!

PR #691, opened 2024-07-16
feat(aws.functions): add conditions in permissions

aryasaatvik:
🚧  WIP


fwang:

PR #693, opened 2024-07-16
sst.aws.postgres: expose node field on PostgresRef

### LINKED ISSUE #4392: Allow getting all nodes of a Postgres construct when using .get()
When using `sst.aws.Postgres..get()` to get a pre-existing PG cluster, it only returns limited info about the cluster.

I think the get should return something as close as possible to instantiating a new Postgres so that the code to either create new or reuse existing Postgres infra can just be a drop in replacement with the the result of the `.get()` fn

I have a PR - [https://github.com/sst/ion/pujll/693](https://github.com/sst/ion/pull/693) - that (partially) addresses this

shmuli9:
Allows access to instance as well as cluster when `get`-ting an existing PostgresDB

### DIFF
--- a/platform/src/components/aws/postgres.ts
+++ b/platform/src/components/aws/postgres.ts
@@ -356,8 +356,8 @@ export class Postgres extends Component implements Link.Linkable, AWSLinkable {
   public static get(
     name: string,
     args: {
-      cluster: rds.GetClusterOutputArgs;
-      instance?: rds.GetInstanceOutputArgs;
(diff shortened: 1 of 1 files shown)


jayair:
Is there an issue related to this? If not can you create one and I'll get the team to review.

shmuli9:
> Is there an issue related to this? If not can you create one and I'll get the team to review.

Created it here - https://github.com/sst/sst/issues/4392

PR #709, opened 2024-07-18
fix: sending cookies in regional remix server with polyfills

luke-rucker:
the fix to sending multiple cookies merged yesterday #702 does not work when remix polyfills are installed. they are missing the `getSetCookie` method amongst a few other issues detailed [here](https://github.com/remix-run/remix/issues/7067). this fixes that


platform/functions/react-server/regional-server.mjs
```diff
@@ -127,6 +127,21 @@ const createApigHandler = (build) => {
...
+const accumulateCookies = (headers) => {
+  // node >= 19.7.0 with no remix fetch polyfill
+  if (typeof headers.getSetCookie === "function") {
+    return headers.getSetCookie();
+  }
+  // node < 19.7.0 or with remix fetch polyfill
+  const cookies = [];
+  for (let [key, value] of headers.entries()) {
+    if (key === "set-cookie") {
+      cookies.push(value);
```
alfredsgenkins, line 139: Should not cookies be accumulated in the following format?

```
`${key}=${value}`
```

Based on this [MDN example](https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie) it should be.

```js
headers.getSetCookie();
// Returns ["name1=value1", "name2=value2"]
```

AWS appears to expect the cookies in `key=value` format, as seen from [this example](https://docs.aws.amazon.com/lambda/latest/dg/services-apigateway.html#services-apigateway-apitypes):

```
"cookies": [
  "s_fid=7AABXMPL1AFD9BBF-0643XMPL09956DE2",
  "regStatus=pre-register"
],
```

platform/functions/remix-server/regional-server.mjs
```diff
@@ -130,6 +130,21 @@ const createApigHandler = (build) => {
...
+const accumulateCookies = (headers) => {
+  // node >= 19.7.0 with no remix fetch polyfill
+  if (typeof headers.getSetCookie === "function") {
+    return headers.getSetCookie();
+  }
+  // node < 19.7.0 or with remix fetch polyfill
+  const cookies = [];
+  for (let [key, value] of headers.entries()) {
+    if (key === "set-cookie") {
+      cookies.push(value);
```
alfredsgenkins, line 142: Same case here.

alfredsgenkins (changes requested):

platform/functions/react-server/regional-server.mjs
luke-rucker: so yes, but in this case the key is not the name of the cookie, but rather the key of the header. the value will contain the whole cookie header

PR #726, opened 2024-07-19
fix: docs for autodeploy event types

nhamilton1:
Fixing the examples for autodeploy. `event.type === "pushed"` is not an option, it should be `event.action === "pushed`. 
`event.type === "pushed" && event.branch === "main"` will change to `event.type === "branch" && event.branch === "main" && event.action === "pushed"`


jayair:
Good catch! Thanks.

PR #743, opened 2024-07-23
Fixes aws monorepo example

dragonautdev:
- Updates references to latest sst package
- Adds sst as a dependency on all projects


jayair:
Thanks.

PR #756, opened 2024-07-24
Adds support for Queue ReportBatchItemFailures

### LINKED ISSUE #755: No support for partial failures on Queue handlers
When your Lambda function encounters an error while processing a batch, all messages in that batch become visible in the queue again by default, including messages that Lambda processed successfully. As a result, your function can end up processing the same message several times.

To avoid reprocessing successfully processed messages in a failed batch, you can configure your event source mapping to make only the failed messages visible again. This is called a partial batch response. To turn on partial batch responses, specify ReportBatchItemFailures for the [FunctionResponseTypes](https://docs.aws.amazon.com/lambda/latest/api/API_UpdateEventSourceMapping.html#lambda-UpdateEventSourceMapping-request-FunctionResponseTypes) action when configuring your event source mapping. This lets your function return a partial success, which can help reduce the number of unnecessary retries on records.

### LINKED ISSUE #4379: No support for partial failures on Queue handlers
When your Lambda function encounters an error while processing a batch, all messages in that batch become visible in the queue again by default, including messages that Lambda processed successfully. As a result, your function can end up processing the same message several times.

To avoid reprocessing successfully processed messages in a failed batch, you can configure your event source mapping to make only the failed messages visible again. This is called a partial batch response. To turn on partial batch responses, specify ReportBatchItemFailures for the [FunctionResponseTypes](https://docs.aws.amazon.com/lambda/latest/api/API_UpdateEventSourceMapping.html#lambda-UpdateEventSourceMapping-request-FunctionResponseTypes) action when configuring your event source mapping. This lets your function return a partial success, which can help reduce the number of unnecessary retries on records.

brunocleite:
Fixes sst/sst#4379


brunocleite:
I have tested with a queue subscription and it activates [ReportBatchItemFailures]

image

Notice that I have enforced the `reportBatchItemFailures` boolean parameter to true, if provided.
This is because I couldn't find a way to accept false as well.

If `false` is accepted for `reportBatchItemFailures` parameter, it must be treated on `apply`. But Pulumi's `functionResponseTypes` demands an array type, not allowing anything else, and passing an empty array fails the deployment. 

The solution found was to lock `reportBatchItemFailures` to true only.

fwang:
Thanks @brunocleite!

Also added batch size and window.

PR #761, opened 2024-07-26
fix: auth sst.cloudflare with email/key pair

### LINKED ISSUE #760: bug: js sdk doesn't work with cloudflare_email and cloudflare_api_key
https://github.com/sst/ion/blob/afae79b33343e9b1a68bd19cebce3d71fbe9dccd/pkg/project/provider/cloudflare.go#L56
^^ says to use cloudflare_api_token or email + api key

### LINKED ISSUE #4376: bug: js sdk doesn't work with cloudflare_email and cloudflare_api_key
https://github.com/sst/ion/blob/afae79b33343e9b1a68bd19cebce3d71fbe9dccd/pkg/project/provider/cloudflare.go#L56
^^ says to use cloudflare_api_token or email + api key, but when I use sst.cloudflare.dns() or something like that, it doesn't authenticate correctly

this happens because this helper function only authenticates with an api token, when the go cli says it supports both email/key and tokens
https://github.com/sst/ion/blob/7ac17cf1c35d2e9a977dd542f5cb8a256a330e7f/platform/src/components/cloudflare/helpers/fetch.ts#L34

rgodha24:
fixes sst/sst#4376 

the `any` on the getAuth() function's return type is because undici's types are being weird

### DIFF
--- a/platform/src/components/cloudflare/helpers/fetch.ts
+++ b/platform/src/components/cloudflare/helpers/fetch.ts
@@ -4,6 +4,9 @@
 const CLOUDFLARE_API_BASE_URL =
   process.env.CLOUDFLARE_API_BASE_URL ?? "https://api.cloudflare.com/client/v4";
 const CLOUDFLARE_API_TOKEN = process.env.CLOUDFLARE_API_TOKEN;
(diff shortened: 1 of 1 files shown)


rgodha24:
@thdxr sorry for pinging you on like everything but can you please review this whenever you get a chance

PR #770, opened 2024-07-28
Json unmarshal error - Update svelte-kit template preset.json

ekaya97:
The comma in line 26 is causing "invalid character ']' looking for beginning of value" error when trying to set up sst@ion with svelte-kit.


jayair:
Thanks!

PR #771, opened 2024-07-28
Added directive for intellisense to find sst module in infra files

NicoPowers:
When using VSCode or similar IDE's with Intellisense, they have trouble finding `sst` inside the `infra` folder and there is no way to import `sst` as a module.

By placing `/// <reference path="../.sst/platform/config.d.ts" />` at the top of the `index.ts` in the `infra` folder, this allows intellisense find `sst` for those files.


Paliago:
I can confirm this solves it for me as well.

jayair:
Can you open an issue for this, I'll ask the team to test it.

thdxr:

PR #783, opened 2024-07-30
update install script for bourne shell

eschaefer:
Bourne shell (`sh`) case is not handled, but has same config behavior as alpine linux `ash`


jayair:
thanks! Can you open an issue for this? I'll assign it to the team.

PR #787, opened 2024-07-31
bug: Vector.QueryResponse type error

### LINKED ISSUE #4358: bug: Vector.QueryResponse type error
The sdk's `QueryResponse` type is Inconsistent with the implemented return type of `VectorClientResponse.query()`.

## Repro

Given:
```
const vectorClient = VectorClient("Vector");
const result = vectorClient.query({ vector, include: { type: "some-identifier" } });
```

Typescript expects `result` to be of type `QueryResponse`. 
Actual result is (correctly) an object containing an Array of 'results'.

sdk's `VectorClient.query` type signature: 
https://github.com/sst/ion/blob/903617e2946d4486ae10c409b5ccae81669f41e9/sdk/js/src/vector/index.ts#L204
query fn implemented return type: 
https://github.com/sst/ion/blob/903617e2946d4486ae10c409b5ccae81669f41e9/platform/functions/vector-handler/index.ts#L93-L98

isolated repro here: https://github.com/jacob-burgess/sst-vector-types-repro

jacob-burgess:
Fixes: https://github.com/sst/sst/issues/4358

## Fix
Update the `QueryResponse` type to match the implemented return type of `VectorClientResponse.query()`.

Likely want to derive the QueryResponse type from the implementation but this is sufficient.

### DIFF
--- a/sdk/js/src/vector/index.ts
+++ b/sdk/js/src/vector/index.ts
@@ -169,13 +169,18 @@ export interface RemoveEvent {
 
 export interface QueryResponse {
   /**
-   * Metadata for the event that was provided when storing the vector.
+   * List of result objects matching the query.
    */
-  metadata: Record<string, any>;
-  /**
-   * The similarity score between the prompt and the queried vector.
-   */
-  score: number;
+  results: {
+    /**
+     * Metadata for the event that was provided when storing the vector.
+     */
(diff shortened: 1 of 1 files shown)


fwang:
Thanks @jacob-burgess!

PR #797, opened 2024-08-01
Fixes outdir naming in the astro example

dragonautdev:
astro-sst is looking for the built astro site inside the `dist` folder, but the docs say you should setup astro to use the `server` folder for builds.


jayair:
Thanks!

PR #736, opened 2024-07-21
remix: derive output directories from vite config

klaemo:
Fixes sst/sst#4437
Fixes sst/sst#4427 
Closes sst/ion#638

This makes the following improvements for remix's vite build:
- detects the remix output directories from the user's vite config. This way we should always get the correct directories, no matter what remix or the user do.
- correctly sets the directory for versioned assets, so we get the correct cache-control headers
- detects`.mts` and `.mjs` versions of the vite config file
- guards against custom values of `serverBuildFile`. We could handle this, if needed.

~~There is a little hack in there to externalize the `vite` import from the esbuild bundle. Otherwise it won't work. I think this should rather be done by defining it as an external package [here](https://github.com/sst/ion/blob/b3a9dcc5cfe05095a25f158c63bf4aa00ad59b24/pkg/js/js.go#L56). But I don't know how to get the whole golang thing including all the JS bits and pieces to build.~~ Fixed this and built the go package.

PR #850, opened 2024-08-13
remix: fix vite config resolution in monorepos

luke-rucker:
this fixes the changes introduced in #736 in monorepos. in both the calls to `vite.loadConfigFromFile` and `vite.resolveConfig`, vite was resolving the root of the site as the cwd of the sst process, not the actual location of the site.

### DIFF
--- a/platform/src/components/aws/remix.ts
+++ b/platform/src/components/aws/remix.ts
@@ -461,12 +461,15 @@ export class Remix extends Component implements Link.Linkable {
           const vite = await import("vite");
           const viteConfig = await vite.loadConfigFromFile(
             { command: "build", mode: "production" },
-            file,
+            path.join(sitePath, file),
           );
           if (!viteConfig) throw new Error();
 
           resolvedConfig = (await vite.resolveConfig(
-            viteConfig.config,
+            // root defaults to process.cwd(), which will be where the sst.config.ts file is located
+            // since we're invoking vite programatically. In a monorepo, this is likely incorrect, and
+            // should be the defined sitePath.
+            { ...viteConfig.config, root: viteConfig.config.root ?? sitePath },
             "build",
             "production",
           )) as Awaited<ReturnType<typeof vite.resolveConfig>> & {


markhker:
I tested this fix and it's working. There is currently and error that prevents deployment in monorepos.

klaemo:
Could catch! I hadn't considered monorepos.

platform/src/components/aws/remix.ts
```diff
@@ -461,12 +461,15 @@ export class Remix extends Component implements Link.Linkable {
           const vite = await import("vite");
           const viteConfig = await vite.loadConfigFromFile(
             { command: "build", mode: "production" },
-            file,
+            path.join(sitePath, file),
```
klaemo, line 464: There's a 3rd argument to `loadConfigFromFile` called `configRoot` (https://vitejs.dev/guide/api-javascript.html#loadconfigfromfile) which I guess could be used here? So:

```ts
vite.loadConfigFromFile({ command: "build", mode: "production" }, file, sitePath)
```

Maybe this also sets `viteConfig.root` to the correct value at the same time.

klaemo: Nevermind, just tried this and the `configRoot` parameter doesn't do anything as far as I can tell

ironheart122:
This fix doesn't work for me. I'm on SST v3.0.42.

image

mattkinnersley:
> This fix doesn't work for me. I'm on SST v3.0.42.

@ironbyte I had the same but realised I was still running an old version. A couple things to check:
- Run `sst version`
- Check `package.json` files and make sure `sst` is not pointing at `ion`.
- Run an `sst install` and `sst refresh` for good measure

ironheart122:
> > This fix doesn't work for me. I'm on SST v3.0.42.
> 
> @ironbyte I had the same but realised I was still running an old version. A couple things to check:
> 
> * Run `sst version`
> * Check `package.json` files and make sure `sst` is not pointing at `ion`.
> * Run an `sst install` and `sst refresh` for good measure

Hey! It was a bit complicated. It turns out the issue was with the remix-flat-routes library. Long story short, I had to make the change shown in my vite config file below:

```ts
import path from 'path'
import { vitePlugin as remix } from '@remix-run/dev'
import { flatRoutes } from 'remix-flat-routes'
import { defineConfig } from 'vite'
import tsconfigPaths from 'vite-tsconfig-paths'

export default defineConfig({
	plugins: [
		remix({
			ignoredRouteFiles: ['**.*'],
			routes: async (defineRoutes) => {
				return flatRoutes('routes', defineRoutes, {
					ignoredRouteFiles: [
						'.*',
						'**/*.css',
						'**/*.test.{js,jsx,ts,tsx}',
						'**/__*.*',
						'**/*.server.*',
						'**/*.client.*',
					],
					appDir: path.join(__dirname, 'app'), // Fixed this
				})
			},

			future: {
				v3_fetcherPersist: true,
				v3_relativeSplatPath: true,
				v3_throwAbortReason: true,
			},
		}),
		tsconfigPaths(),
	],
})

```

PR #852, opened 2024-08-14
docs(astro-sst): update wrong astro output

unrevised6419:
The only valid astro output values are 'static' | 'server' | 'hybrid'

image

### DIFF
--- a/www/src/content/docs/docs/start/aws/astro.mdx
+++ b/www/src/content/docs/docs/start/aws/astro.mdx
@@ -43,7 +43,7 @@ It'll also ask you to update your `astro.config.mjs` with something like this.
 + import aws from "astro-sst";
 
 export default defineConfig({
-+  output: "dist",
++  output: "hybrid",
 +  adapter: aws()
 });
 ```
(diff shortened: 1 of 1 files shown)


jbrown:
I had the same issue after following the Astro guide. Check this commit from 2 weeks ago:

https://github.com/sst/ion/commit/4f55966b36f4854f5a256d8c46d2e3e051dab919

@jayair is there an explanation for this change?

The [README for astro-sst](https://github.com/sst/sst/tree/master/packages/astro-sst) still says to put `output: "server"`

jayair:
Oh man somebody opened an issue about that change and I put it in. Now I can't find it.

@iamandrewluca can you change this back to "server" then.

www/src/content/docs/docs/start/aws/astro.mdx
```diff
@@ -43,7 +43,7 @@ It'll also ask you to update your `astro.config.mjs` with something like this.
 + import aws from "astro-sst";
 
 export default defineConfig({
-+  output: "dist",
++  output: "hybrid",
```
unrevised6419:
```suggestion
+  output: "server",
```

jayair:
Actually, let me check with @fwang because this what we do internally:

https://github.com/sst/ion/blob/dev/www/astro.config.mjs#L177

unrevised6419:
@jayair too late, changed to `server` already ))

unrevised6419:
Fixed in https://github.com/sst/ion/commit/6e3b26b03a4e9d9432de0ccdc622fb8e81551b95

PR #864, opened 2024-08-15
docs(live.mdx): adding Live resource dependency

ViniciusAlberkovics:
Added list of resources that live mode depends on.
Mapped fix for infinite "Deploying" problem in case of any manual resource deletion.


jayair:
Thanks! But I'm going to move this to a separate doc once I get to this.

jayair:
Moved it to the State doc https://github.com/sst/ion/commit/336dc8b1aef2087f02e38cb69d146e0baa11f37f

PR #878, opened 2024-08-18
fix: specific version installation for 3.0

### LINKED ISSUE #4269: Specific version installation is not working for 3.0
Currently running the following gives an error
```bash
curl -fsSL https://ion.sst.dev/install | VERSION=3.0.42 bash
```
image

 This fails because the script always uses 'v' prefix when downloading, but for 3.0+ there is no 'v' prefix.

nurulhudaapon:
Currently `curl -fsSL https://ion.sst.dev/install | VERSION=3.0.42 bash` fails because the script always uses 'v' prefix when downloading, but for 3.0+ there is no 'v' prefix.

Fixes: [879](https://github.com/sst/sst/issues/4269)

### DIFF
--- a/install
+++ b/install
@@ -48,6 +48,9 @@ if [ -z "$requested_version" ]; then
     fi
 else
     url="https://github.com/sst/ion/releases/download/v${requested_version}/$filename"
+    if [ "${requested_version:0:1}" -ge 3 ]; then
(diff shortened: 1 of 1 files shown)


jayair:
Sounds like Dax fixed this?

I'll close this. Feel free to reopen if it's still broken.

PR #898, opened 2024-08-21
Vector: static get method

### LINKED ISSUE #897: Vector unable to share Postgres cluster across stages
The Vector class does not have a static `get` method. 

Without this, a new Postgres cluster must be created for each stage.

PR: #898

### LINKED ISSUE #4253: Vector unable to share Postgres cluster across stages
The Vector class does not have a static `get` method. 

Without this, a new Postgres cluster must be created for each stage.

PR: sst/ion#898

jacob-burgess:
fixes sst/sst#4253

Implement a static get method on the Vector component so that the resource can be shared between stages.

## Implementation

When instantiating a new Vector via the `get` method, the constructor `get`s the existing Postgres cluster via the clusterID rather than creating a new cluster.

Note: Each stage still creates its own lambda functions to interface with the DB. Dont see why that would be a problem, but calling out bc slightly different pattern than other implementations of `static get`


jacob-burgess:
Taking second pass - maybe for these higher level components, this sort of thing belongs in the transform? Looking at what the `sst.config.ts` would look like to share a Vector between stages... 

As currently implemented, would look like:
```ts title="sst.config.ts"
const vector =
  $app.stage === "dev"
    ? new sst.aws.Vector("MyVector", {
        dimension: 1536,
      })
    : sst.aws.Vector.get("MyVector", "cluster-id");
```
note: no "new" keyword - but we _are_ creating a new Vector (at least the lambdas to interact with it)?

Using `transform` might look something like this? (this currently type errors)
```ts title="sst.config.ts"
const vector2 =
  $app.stage === "dev"
    ? new sst.aws.Vector("MyVector", {
        dimension: 1536,
      })
    : new sst.aws.Vector("MyVector", {
        dimension: 1536,
        transform: {
          postgres: sst.aws.Postgres.get("MyVector", "cluster-id"),
        },
      });
```
> Type 'Postgres' is not assignable to type 'Transform<PostgresArgs> | undefined'

I haven't gotten the transform to work, maybe im dumb. But something like this _feels_ a bit better to me (and maybe the whole point of transform). Like "create a new Vector, but btw use the same underlying postgres instance"

it doesnt exactly fit - will see ya'lls thoughts before I go yucking up all the vector stuff

fwang:
Thanks @jacob-burgess!

Using `.get()` makes sense. Ideally the lambda functions should also be referenced. Perhaps the functions should be tagged w/ the cluster id, so they can be looked up.

PR #900, opened 2024-08-21
feat(realtime): Add the ability to provide custom authorizer fields

ian-pascoe:
Updated the docs to include an example of providing one of the new options: `principalId`.

This opens the door for adding shadows to the IoT device including auth data via the principalId.


ian-pascoe:
This resolves sst/sst#4252

PR #902, opened 2024-08-21
aws.Service: place ECS tasks in public subnet if cluster VPC has no NATs

Dizzzmas:
Issue: https://github.com/sst/sst/issues/4251

## What does this do?

- Normalizes `args.vpc` used in `aws.Service` to have the interface of SST's `Vpc` component so that more properties are available
- If `vpc.nodes.natGateways` is empty - deploy ECS tasks into public subnet

## How was this tested?

Published sst package locally with [yalc](https://github.com/wclr/yalc)

Then deployed below example and made sure the public subnets were used
```typescript
    const vpc = new sst.aws.Vpc("TestClusterVpc", {
      nat: false,
    });
    const cluster = new sst.aws.Cluster("TestECSCluster", { vpc });
    cluster.addService("TestClusterService", {
      image: ".",
      public: {
        ports: [
          { listen: "80/http" },
        ]
      },
    })

```

### DIFF
--- a/platform/src/components/aws/service.ts
+++ b/platform/src/components/aws/service.ts
@@ -141,23 +141,12 @@ export class Service extends Component implements Link.Linkable {
     function normalizeVpc() {
       // "vpc" is a Vpc component
       if (args.vpc instanceof Vpc) {
-        const result = {
-          id: args.vpc.id,
-          publicSubnets: args.vpc.publicSubnets,
-          privateSubnets: args.vpc.privateSubnets,
-          securityGroups: args.vpc.securityGroups,
-        };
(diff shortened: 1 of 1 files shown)


fwang:
Thanks @Dizzzmas!

We just released a new version of Cluster that deploys to the public subnet by default.

tday:
@fwang, do existing clusters need to do anything to avoid being deployed to a public subnet since it's the new default?

fwang:
The component will not auto update. On ur next deploy, it should error with instructions to change `sst.aws.Cluster` to `sst.aws.Cluster.v1` to stay on the previous cluster setup.

PR #904, opened 2024-08-21
ApiGatewayV2: Pass provider to subcomponents

dominiklessel:
See https://discord.com/channels/983865673656705025/983865673656705028/1275833084276637767

## What does this do?

This PR ensures that the provider configuration is properly passed down to the subcomponents. Currently this isn't the case and will cause problems if you try to deploy an ApiGatewayV2 to a different region than the main stack.

CleanShot 2024-08-21 at 17 03 05@2x

## How was this tested?

Changed the files locally in the `.sst` folder.

Then deployed the example below and made sure it worked:

```ts
const usEast1 = new aws.Provider("usEast1", { region: "us-east-1" });

const someApiThingy = new sst.aws.ApiGatewayV2(
  "someApiThingy",
  undefined,
  { provider: usEast1 }
);

someApiThingy.routeUrl(
  "ANY /{proxy+}",
  "https://api.something.com/{proxy}"
);

someApiThingy.route("GET /foo", "apps/api/src/handlers.index");

export const proxies = new sst.Linkable("Proxies", {
  properties: {
    someApiThingyUrl: someApiThingy.url,
  },
});

```

### DIFF
--- a/platform/src/components/aws/apigateway-websocket.ts
+++ b/platform/src/components/aws/apigateway-websocket.ts
@@ -197,6 +197,7 @@ export interface ApiGatewayWebSocketRouteArgs {
 export class ApiGatewayWebSocket extends Component implements Link.Linkable {
   private constructorName: string;
   private constructorArgs: ApiGatewayWebSocketArgs;
+  private constructorOpts: ComponentResourceOptions;
   private api: apigatewayv2.Api;
   private stage: apigatewayv2.Stage;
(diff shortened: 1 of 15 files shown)


fwang:
Thanks @dominiklessel! I reverted the change for `ApiGatewayV2Authorizer` and other component. `provider` is automatically inherited from `parent`.

dominiklessel:
Cool, thank you so much!

PR #931, opened 2024-08-25
feat: Python functions

walln:
I implemented the Python runtime for both deployed functions and live lambda.  It uses [uv](https://github.com/astral-sh/uv) to handle local development dependencies and the Python runtime. ~~Live dev works, but I am still working on capturing logs locally.~~

~~It still needs some more work, such as documentation and fixing logs, but I wanted to put this out there for feedback.~~

Also, excuse my go code; I'm not very experienced with it.


walln:
Live dev with logs and dependencies works. SST will now install UV, so you do not even need to have Python installed to start using SST with Python functions. The only caveat is that since deps are not bundled/compiled, you must consider the 50 MB lambda limit when using bloated packages.

jayair:
Oohhh

walln:
I'm curious if anyone, especially the core team, has thoughts on having Python runtimes deploy with containers rather than the native Python runtime. This makes sense to me for a couple of reasons:

Python is commonly used with packages that have low-level system dependencies, such as the scipy ecosystem. This can introduce the need for binaries built for specific os/arch, complexities with building deps, etc. Using containers would enable users to hook into the build process and ensure their production builds work as expected.

For larger deployments (pretty typical with Python lambdas), containers have better performance [see here](https://aaronstuyvenberg.com/posts/containers-on-lambda#:~:text=Lambda%20recently%20improved%20the%20cold,images%20and%20you%20should%20too.), and containers would make the build process a bit less of a hack. I am surgically copying the dependencies from a virtualenv into the deployed zip, but this is likely to have some issues with specific dependencies, have platform-specific problems, and be less extensible.

Before moving this to deploy to containers in prod, I wanted to get feedback and some insight into any future container plans I should consider if I proceed with containers.

CestDiego:
I love the containers idea but in aws it gets very expensive very soon with the need for nat and ecr :( are there alternative ways to not have such high cost ?

walln:
I agree that paying the ECR cost on top of lambda usage is less than ideal ($0.1/Gb/mo), but it is still the best route to enable the most flexibility while providing a great out-of-the-box experience. The default container image will be as slim as possible to mitigate the cost and cold start times.

Regarding NAT, that should only be added pricing if you are adding your functions to a VPC which is not the default for SST functions.

thdxr:
i think what makes sense is to have a `runtime: "python"` and a `runtime: "python_docker"` - the latter would automatically do all the work to ship your code as a docker image without you having to create the Dockerfile

although - is that possible? if so i think an opinionated setup around that makes sense - eventually we'll have `runtime: "docker"` for general docker functions for people who want to deploy a custom image

walln:
I have been thinking about this today and think I've settled on something good. Rather than having the runtime as `runtime: "python"` or `runtime: "python_docker"`, it might be better to have a python-specific argument similar to how `sst.aws.Function` has `nodejs` options—something like: `python: { container: bool }` to indicate the deployment target. 

And I see no reason it shouldn't be possible. SST can ship dockerfiles like it has the runtime bridges/servers; during the build step, copy the dockerfile into the artifact directory for the function, perform the build, upload it to ECR, and deploy it. If the user has a dockerfile in the handler's directory, use that instead. 

For the future `runtime: "docker"` you would probably want the Function args to explicitly specify a path to a dockerfile as you probably intend a more custom container definition. But the Python build is pretty simple in the dockerfile: copy PyProject and lockfile (uv lockfiles are cross-platform - thank you, Astral!), install the deps and copy the function code. Having SST-supported runtimes that deploy to containers configured under the runtime's namespace is better because it makes it clearer that the user does not have to own the docker file.

I am thinking something like this:
```ts sst.config.ts
new sst.aws.Function("MyPythonFunction", {
    handler: "./functions/python-api.handler",
    runtime: "python",
    python: {
        container: true
    }
}
```
```
project-root/
├── functions/
│   ├── pyproject.toml
│   └── python-api.py
└── sst.config.ts
```
SST will automatically dockerize everything in the same directory of the handler function (and the closest parent `pyproject.toml` and lockfile) and deploy it; nothing is needed from the user. But if a dockerfile is in the directory, use that instead. I like this design because it also makes sense for future first-party runtimes that should not require an upfront dockerfile, e.g., Deno/Bun. I could also have a good chunk of the logic that can be reusable for future runtimes. Also, if SST owns the default dockerfile, we can automatically bundle an SST SDK just like it is done in typescript functions; I have some ideas on how to do this, but I need to experiment because Python local package resolution can be a bit strange.

The explicit container runtime should clearly state that the user owns its configuration. For example, an exciting use case is deploying external software using lambda as the deployment platform, like some saas pre-existing container

```ts sst.config.ts
new sst.aws.Function("PaymentService", {
    runtime: "docker",
    docker: {
        dockerfile: "./path/to/my/dockerfile"
    }
}
```
```dockerfile
# assume this exists
FROM cool-saas:lambda:whatever
CMD ["my", "lambdaric", "compatible", "args"]
```
The explicit container runtime is beyond the scope of this PR and will likely not leverage the logic from the first-party container deployments, but I could look into supporting it in a follow-up or later.

TLDR:  Container runtimes are either first-party, such as Python, where SST provides a default container, or `docker`, where the user must supply the dockerfile/image. The same Python runtime will have a "deploy to container" mode that requires no changes other than setting `python: { container: true }` for your function. Live-dev and dependency configuration will work as currently implemented.

I have a minimal version kind of working and will probably have something more complete later this week.

valgaze:
Ancient history at this point but worth a look for inspiration: 

https://github.com/simonireilly/sst-python-api

https://github.com/simonireilly/sst-python-api/blob/main/api-stack/lib/MyStack.ts

walln:
I got container support working. 0 changes are needed, just a single `python: { container: true }` Live dev, uv dependency management, etc, and an example with numpy deployed. I still need to think about the SDK ideas, but I made the container support very extensible; it should be as simple as defining a build function for the specific runtime and a base dockerfile; the container building and deploying stuff are generic.

thdxr:
yep this api is better than what i suggested - really like it! makes sense for docker to just be an option. let me know if you feel like it's at a ready point and i'll do some testing on my end and get it merged

walln:
Python has duck typing and a not-so-great type inference system, so it probably is not necessary, but I wanted to create the same type of generation the JS SDK gets if that's something the SST team is interested in. You just need to publish the SDK similarly to the JS SDK, if any sst.aws.Function is using the python runtime a sst_env.pyi file is generated at the root that users can include as a type hint file in their editor. If this is too much, I can remove it as well since I imagine the SST team does not envision Python as a "Tier 1 support" language.

```python
# --- sst_env.pyi
# Automatically generated by SST
# pylint: disable=all
from typing import Any

class Resource:
    class MyPythonFunction:
        name: str
        type: str
        url: str



# --- my_handler.py
from sst_sdk import Resource

def handler(evt, ctx):
    print(Resource.MyBucket.Name)
    return {
        "statusCode": 200,
        "body": f"Hello World",
    }
```

walln:
I added an example using huggingface transformers, I'm sure this is something some people would be interested in; it also shows how to resolve custom Pypi links for nonstandard packages such as Pytorch.

```py
from transformers import pipeline
import json

# Initialize the pipeline once (outside the handler) to avoid reloading the model on each request
pipe = pipeline("text-generation", model="roneneldan/TinyStories-33M")


def handler(event, context):
    # Define the prompt for text generation
    prompt = "Write a short story about the magical framework SST that makes the cloud so easy to use!"

    # Generate text using the pipeline
    response = pipe(
        prompt,
        max_length=150,
        num_return_sequences=1,
        temperature=0.6,
        top_k=50,
        do_sample=True,
    )

    # Extract the generated text
    generated_text = response[0]["generated_text"]

    # Return the response with a status code and JSON-encoded body
    return {
        "statusCode": 200,
        "body": json.dumps({"story": generated_text}),
    }
```

thdxr:
very cool - yeah i need to clean up our type generation stuff, have some changes in a branch but want to make it a bit more modular so we can add support for other targets

sommeeeer:
this is awesome!

thdxr:
updated this PR for the new runtime structure

this PR is very very good - probably the best one we've ever received.

only thing that's blocking is one question - what do you think of not installing `uv` always and expecting the user to have it installed?

we try to only install things globally if we know for sure all sst users will need it and since uv is only required for python functions i think it might make sense to ask the user to install it

does that make sense?

walln:
That's some high praise!

I thought about that but was not sure what the thinking was regarding binary dependencies. What do you think about centralizing the binary deps in `pkg/global/global.go` to make it so that when the `<BINARY_NAME>Path()` function is called, it will install the dependency if it does not exist? I imagine that in the future, SST will need to add more and more binary dependencies (go and rust functions come to mind, not sure about docker?), I think having SST install everything it needs makes sense, especially for cases like CI deployments.

thdxr:
yeah that's a good call let me see if i can make it lazy loadable

btw i'm doing some rework to do _all_ function building in go so that we don't need to make two implementations in js and in go but i'll take care of porting

subwaymatch:
This is amazing. Thank you @walln for the PR. I can't wait to use the new API . 👍

fverbist:
@walln awesome! Will be trying it soon...
I'm also interested in adding java functions. 
would you give me a couple pointers? I want to give it a shot.

PR #960, opened 2024-08-30
rust function url example

judegiordano:
an example deploying a rust api on a function url


jayair:
oh wow nice, I'll try this next week

PR #964, opened 2024-08-31
bugfix: minor config fix for astro guide

joshymcd:
minor fix for the astro site setup guide


jayair:
Yeah I need to figure this out on our side

https://github.com/sst/sst/issues/4244

jayair:
Yeah I ended up fixing it elsewhere cos I had to update the preset for the template as well. Thanks!

PR #968, opened 2024-09-01
static-site: add support for custom viewer request handler

### LINKED ISSUE #934: StaticSite with Custom CloudFront Function
I am migrating one of our legacy Angular app to CloudFront + S3. As we use an old version of Angular's i18n package, the output is separated per language, resulting in the following structure:

```
/en/index.html
/fr/index.html
```

I need to implement the following behavior:
- example.com/en/* -> en/index.html
- example.com/fr/* -> fr/index.html
- else -> 301 -> example.com/en/$1

This can be done with the CloudFront Function, which cannot be modified at the moment (or I am blind - totally possible).

Do you plan to support this anytime soon?

### LINKED ISSUE #4226: StaticSite with Custom CloudFront Function
I am migrating one of our legacy Angular app to CloudFront + S3. As we use an old version of Angular's i18n package, the output is separated per language, resulting in the following structure:

```
/en/index.html
/fr/index.html
```

I need to implement the following behavior:
- example.com/en/* -> en/index.html
- example.com/fr/* -> fr/index.html
- else -> 301 -> example.com/en/$1

This can be done with the CloudFront Function, which cannot be modified at the moment (or I am blind - totally possible).

Do you plan to support this anytime soon?

balintant:
Adds `viewerRequestFunction` to aws/StaticSite to customize the CloudFront Function (Closes sst/sst#4226).

(This was a quick POC that I can temporarily use instead of implementing the same with Terraform. Please let me know what needs to be changed to get this into a release.)


fwang:
Thanks @balintant, i tweaked the api a bit to take a pre-created CF function. This allows for better composability. Down the road we could add a CloudFront Function component that helps ppl bundle and minimize the code.

Here's an example:
```ts
    const fn = new aws.cloudfront.Function("fn", {
      runtime: "cloudfront-js-2.0",
      // redirect all requests to google.com
      code: `
      function handler(event) {
        var response = {
          statusCode: 302,
          statusDescription: 'Found',
          headers: {
            'location': { value: 'https://www.google.com' }
          }
        };
        return response;
      }
      `,
    });

    new sst.aws.StaticSite("Web", {
      edge: {
        viewerRequest: fn.arn,
      },
    });
  },
```

PR #975, opened 2024-09-02
Update set-up-a-monorepo.md

joneslloyd:
Per: https://github.com/sst/sst/issues/4191


jayair:
Thanks but this isn't specific to the monorepo setup, I'll need to figure out a different spot for it.

PR #982, opened 2024-09-03
sst.aws.ApiGatewayV1: fix issue with linking certain SST components to route handlers

JohnG45:
Fix for https://github.com/sst/sst/issues/4238.

### DIFF
--- a/platform/src/components/aws/apigatewayv1.ts
+++ b/platform/src/components/aws/apigatewayv1.ts
@@ -765,10 +765,27 @@ export class ApiGatewayV1 extends Component implements Link.Linkable {
   ) {
     const { method, path } = parseRoute();
     const prefix = this.constructorName;
-    this.triggers[`${method}${path}`] = jsonStringify({
-      handler,
-      args,
-    });
+    if (typeof handler === "string") {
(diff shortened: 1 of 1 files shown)


platform/src/components/aws/apigatewayv1.ts
```diff
@@ -765,10 +765,27 @@ export class ApiGatewayV1 extends Component implements Link.Linkable {
...
+    } else {
+      this.triggers[`${method}${path}`] = jsonStringify({
+        handler: {
+          ...handler,
+          // If any of the linked SST components have Pulumi resources as instance variables, this code can throw
+          // `TypeError: Converting circular structure to JSON` (see https://github.com/sst/ion/issues/918), as
+          // the Pulumi resource's __parentResource property references the SST component.
+          // This protects against that, whilst still ensuring that a deployment will be triggered if anything
+          // about any of the links changes.
+          link: $output((handler as FunctionArgs).link).apply((links) =>
```
JohnG45, line 782: I do wonder whether we even need to serialise the links here, i.e. does a change to the links mandate an API Gateway stage deployment?

thdxr:
we have a fix in v3.0.99 - turns out was a simpler problem

PR #996, opened 2024-09-06
Custom origins for SSR site plan

rmarscher:
Hi! It would be sweet to be able to add custom origins to the SSR site plan.

I have a component that uses the functions from ssr-site.ts. I accept a "beforePlanValidate" arg which is an optional function hook that gets the plan right before I call validatePlan. Using that combined with this edit to support custom origins allows me to map additional functions and s3 buckets - i.e. a lambda function with my custom image optimizer, another function with API endpoints and an s3 origin for an uploaded media bucket.

I imagine more docs/tests/etc would be needed to properly release this, but I thought I could get the discussion started with an example implementation.

Thanks!

### DIFF
--- a/platform/src/components/aws/ssr-site.ts
+++ b/platform/src/components/aws/ssr-site.ts
@@ -2,6 +2,7 @@ import path from "path";
 import fs from "fs";
 import { globSync } from "glob";
 import crypto from "crypto";
+import { CloudFrontCustomOrigin } from "aws-lambda";
 import {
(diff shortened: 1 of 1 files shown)


jayair:
Can you open an issue for this as well? We'll review it.

rmarscher:
Sorry for the slow reply here. I opened issue sst/sst#4149. Thanks for considering it and thanks for this awesome project!

PR #1007, opened 2024-09-07
auth: allow basePath configuration

### LINKED ISSUE #1065: auth: basePath is not configurable
Because it's not configurable, we can't put the auth handler behind a non-default path of a Router.

### LINKED ISSUE #4123: auth: basePath is not configurable
Because it's not configurable, we can't put the auth handler behind a non-default path of a Router.

santiagomera:
This allows auth to work behind a non-default path of an sst.aws.Router. Fixes sst/sst#4123

### DIFF
--- a/sdk/js/src/auth/handler.ts
+++ b/sdk/js/src/auth/handler.ts
@@ -65,6 +65,7 @@ export function AuthHandler<
     >;
   }[keyof Providers],
 >(input: {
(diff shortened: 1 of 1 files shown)


jayair:
Is there an issue for this? If not, can you create one?

santiagomera:
No, doesn't have one. Will do.

PR #1010, opened 2024-09-07
docs: Correct postgres documentation error

the10xProcrastinator:
Documentation currently states that the maximum ACU range that can be configured on Aurora Postgres serverless v2 is between 0.5 to 128, but the actual range is 1 to 128.

If you try to configure a max ACU of 0.5 you'll get the following error:
```InvalidParameterValue: Serverless v2 maximum capacity 0.5 isn't valid. The maximum capacity must be at least 1.0.```

The range can also be seen on the AWS Console:
image


jayair:
Thanks!

PR #1017, opened 2024-09-09
rework type generation

thdxr:


oscartbeaumont:
Can this be turned off or configured?

I'm getting `sst-env.d.ts` being generated for all of `packages/*` even though none of them can or should touch the infrastructure, only the actual apps within `apps/*` should.

jayair:
@oscartbeaumont create a new issue?

PR #1032, opened 2024-09-10
Update container.mdx

KyleAMathews:
`nat: "managed"` is required now (so Typescript tells me).


jayair:
Oh this example needs an update then. Thanks.

jayair:
Wait hold on, is it? I'm not seeing it.

Which version btw?

PR #1033, opened 2024-09-10
clarify where to run the live command from

zvictor:
After spending time researching is that I realized VS Code requires it to be executed from within their terminal windows.


jayair:
Thanks!

zvictor:
If anyone else is interested, I figured a [hacky way of running things from outside vscode](https://github.com/microsoft/vscode-js-debug/issues/1992#issuecomment-2346183206). I hope it helps!

PR #1043, opened 2024-09-12
feat: include aws-rds-instance example for mysql free tier. with public access enabled

john20xdoe:
Based off AWS RDS Free Tier MySQL
https://aws.amazon.com/rds/free/

Basic configuration.


jayair:
Thanks!

PR #1061, opened 2024-09-13
Update index.mdx

af-lozano:
To access the CLI it must be installed globally. I guess it's obvious to most people familiar with npm but nonetheless.


jayair:
If you are using Node we recommend not installing it globally.

Are you using this for a Node project?

af-lozano:
I was following the guide and it does say to install it locally. But on the dev section it says to run  only `sst dev`. If it's installed locally it won't work because sst wont be on the global node modules and PATH won't pick it up. I think if the recomendation is to install it locally the command on the dev section should be `npx sst dev`.

jayair:
It's a bit tricky because if you are not using Node you should install it globally. But for Node we don't want you to.

Which is why we drop the `npx` part in the docs. 

I'll need to figure something out but closing this for now.

PR #1074, opened 2024-09-16
fix: change Cloudfront Function handler to async function where code injection happens

leoalves:
This Pull Request modifies the Cloudfront function handler to be async in the sections where users can inject custom code. 

This change is necessary to accommodate scenarios where asynchronous operations, such as accessing the Cloudfront KV store or when `await` is required in the user provided code.

### DIFF
--- a/platform/src/components/aws/router.ts
+++ b/platform/src/components/aws/router.ts
@@ -605,7 +605,7 @@ export class Router extends Component implements Link.Linkable {
           runtime: "cloudfront-js-2.0",
           keyValueStoreAssociations: config?.kvStores ?? [],
           code: `
-function handler(event) {
+async function handler(event) {
   ${
     injectHostHeader
       ? `event.request.headers["x-forwarded-host"] = event.request.headers.host;`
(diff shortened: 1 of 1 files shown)


paolostyle:
Thoughts on sst/sst#4126? I believe there will be more cases where the current `injection` pattern will be problematic.

leoalves:
Hello @paolostyle,

Unfornately,  I lack familiarity with the codebase to have an opinion about this. A quick look at the frontend components reveals extensive use of Cloudfront Functions for modifying headers, caching behavior, and other request/response aspects.  I am not sure what would be the unexpected behaviours your suggeston could cause.

IMO, SST should offer the flexibility to pass either the ARN of a CloudFront Function or an object containing the injection code and KV stores.  But again, I am not familiar with the codebase to understand the effects of this change.

It's important to note that this specific PR is unrelated to the decision of how to handle user-provided functions. That decision should be addressed in a separate PR or within the related issue you submitted.

This PR focuses solely on adding `async` to the injection function handler, which appears to be a necessary fix. Since integrating the function with a KV store requires asynchronicity to enable access.

Thanks

fwang:
Thanks @leoalves!

Fixed in https://github.com/sst/ion/commit/2703951f253404f5f1e9afe5b9c82719090b9e1a. Had to change a few places.

Will be in the next release.

leoalves:
Thanks @fwang,

I didn't change in the other places because I didn't think they used the injection pattern and the funciton body didn't have any await calls.  But I think it makes sense to have it everywhere to make it consistent.

I will test when the next release is published.

PR #1082, opened 2024-09-17
fix: python docker build architecture

### LINKED ISSUE #1079: Python function container build targets wrong architecture
When you run `sst deploy` from your local machine, Docker seems to build for your local machine rather than the architecture of the target lambda function.  I'm running an M1 Macbook and when I deployed the Lambda with `sst deploy` it resulted in the lambda not being able to be run.
When the same code was deployed from a machine with the same architecture as the lambda, the lambda worked as expected.

```
{
    live: false,
    handler: "someHandler/main.handler",
    runtime: "python3.11",
    python: {
      container: true
    }
}
```

Maybe an additional flag needs to be passed to docker to ensure its built with the same architecture as the lambda.

### LINKED ISSUE #4115: Python function container build targets wrong architecture
When you run `sst deploy` from your local machine, Docker seems to build for your local machine rather than the architecture of the target lambda function.  I'm running an M1 Macbook and when I deployed the Lambda with `sst deploy` it resulted in the lambda not being able to be run.
When the same code was deployed from a machine with the same architecture as the lambda, the lambda worked as expected.

```
{
    live: false,
    handler: "someHandler/main.handler",
    runtime: "python3.11",
    python: {
      container: true
    }
}
```

Maybe an additional flag needs to be passed to docker to ensure its built with the same architecture as the lambda.

mtmeyer:
Currently when you build a docker image for python lambdas on a machine with a different architecture to the target lambda, the lambda will fail to run.

This PR passes through the architecture the lambda uses to Pulumi's docker build.

Tested this locally and I can now run `sst deploy` from my M1 mac and it works on an `x86_64` lambda.

Closes sst/sst#4115


walln:
Thank you! I totally forgot to add this back after bashing my head against the wall with buildx cross platform images.

fwang:
Thanks @mtmeyer!

PR #1100, opened 2024-09-19
feat: Adde doc to handle the sst deploy for react typescript project

yashwanth2804:


jayair:
Thanks but this is an auto-generated file.

Can you move your docs to this comment block instead? https://github.com/sst/ion/blob/dev/examples/aws-vite/sst.config.ts#L6

fwang:
@thdxr is this still required w/ the recent change in how types are generated?

thdxr:

PR #1113, opened 2024-09-21
Clarify documentation version

scriptcoded:
I just started reading up on SST. I got confused by the tip message since I thought I was on the v1 documentation. This change clarifies which version is documented.

I didn't raise an issue for this, sorry about that! Forking, making the change and opening a PR took no more than 2 minutes, so just reject it if you don't agree.

Cheers,
Malcolm


jayair:
Ah yeah somebody else brought this up. I ended up removing this anyway.

Thanks!

PR #1114, opened 2024-09-21
nuxt: detect Nuxt when using app.config.ts in project

ZainW:
Issue: app.config.ts is used in apps with nuxt ui/ui-pro packages from the nuxt team. they use it to configure certain aspects. when this is here SST thinks it's solidstart

fixed it just by changing the order of the switch case

idk how to really run the project cause i've never used a go project before

Reference: https://ui.nuxt.com/getting-started/theming#appconfigts

### DIFF
--- a/cmd/sst/init.go
+++ b/cmd/sst/init.go
@@ -79,18 +79,18 @@ func CmdInit(cli *cli.Cli) error {
 		template = "astro"
 		break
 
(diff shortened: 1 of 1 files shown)


ZainW:
actually read the Readme properly, ill test it out and report back

ZainW:
nuxt 4 will fix this with the new directory im closing

PR #1118, opened 2024-09-22
feat: add ses identity notification

yyaskriloff:
Assuming most projects are going to production setting up identity notifications is needed to mange complaints and bounce. 

I tried to match the coding style if theres anything you need me to change just lmk and i'll try to do it as soon as i have time.


jayair:
Oh interesting. Can you open an issue for this?

I also think we might want to take the ARN here instead of the ARN @fwang

fwang:
Thanks @yyaskriloff!

Event destinations support more events types and destinations, and is preferred over identity notifications. I re-implemented what you had w/ event destinations here https://github.com/sst/ion/commit/ce169d8fdc29927989bb59e29461fed8cf835f2a

Will be in the next release.

^ feel free to open an issue if this doesn't work for you.

PR #1125, opened 2024-09-23
feat: Add TaskDecleration component

yyaskriloff:
i wanted to add a task definition component in order to use with the Cron component. Being able to schedule tasks that regular lambdas don't work for.

This is the first part and I'm looking for some guidance on how to do this.

1. how abstracted do you want to go? in this i essentially copied the Service component and removed anything thats not relevant. I can remove the logic from Service that the TaskDefinition is doing i just didnt know if you wanted Cluster -> addService -> Service -> createTaskDefintion -> TaskDefintion. I did try. it was taking a while so i thought i can go back if thats how you wanted to do it if this PR is accepted

2. if you think this is a good idea i need to also change the Cron component which means it also has to be able to accept a task definition and run a task. do you want me to create a new component or rework the job arg?


yyaskriloff:
https://www.pulumi.com/registry/packages/aws/api-docs/cloudwatch/eventtarget/#ecs-run-task-with-role-and-task-override-usage

yyaskriloff:
I thought i'de be done by now. I'll create a new pr when i'm finished creating the cron

PR #1091, opened 2024-09-18
fix: python version and sdk resolutions

walln:
Added support for all AWS supported python runtimes except for 3.8 since it gets deprecated in October, and aligned all template pyproject.toml files to ensure local dev is consistent.

Something is wrong with resource linking and its not just a python function issue, waiting on some changes for that potentially, so this is not ready to merge.

PR #1134, opened 2024-09-24
Feat: add copyFiles for container python functions

spoeck:
If you define following python function in your sst.config.ts, the copyFiles are missing in the built container:

```typescript
 const python = new sst.aws.Function("MyPythonFunction", {
      handler: "backend/python.handler",
      runtime: "python3.11", 
      url: true,
      copyFiles: [
        {
          from: "backend/src/mymodule.py",
          to: "backend/src/mymodule.py",
        },
        {
          from: "backend/src",
          to: "src",
        },
        {
          from: "backend/data/sample.csv",
          to: "sample.csv",
        },
      ],
      python: {
        container: true,
      },
    });
```

Now we copy the files defined in copyFiles into the root folder which the Dockerfile later on copies and build the python container with it.


spoeck:
This problem was fixed with #1091 by @walln
I close my PR

PR #1141, opened 2024-09-25
fix: Broken python bundling for nested python functions + add example

### LINKED ISSUE #4114: Python functions fail to build with mono repo structure / nested folders
It seems as though when you have a python function defined as follows

```
    new sst.aws.Function("MyPythonFunction", {
      handler: "projects/function-a/src/logic.handler",
      runtime: "python3.11",
      url: true,
    });
  ```
  
  You'll hit the following error during the build
  
```
|  Error: Failed to build function "projects/function-a/src/logic.handler": Error: ENOENT: no such file or directory, copyfile '/Users/chris/test-project/projects/function-a/pyproject.toml' -> '/Users/chris/test-project/.sst/artifacts/MyPythonFunction-src/Users/chris/test-project/projects/function-a/pyproject.toml'
|      at file:///Users/chris/test-project/.sst/platform/src/components/aws/function.ts:1378:23 {
|    promise: Promise { <rejected> [Circular *1] }
|  }
```

Guessing this needs to trim the target path to be of the same parent as the project file
https://github.com/sst/ion/blob/5b183e67ed7b9134cd05577b2d417fc7d482534b/platform/src/runtime/python.ts#L73C1-L76C7

cmanou:
- Add the example to show the broken issue 
- Turns out findAbove either returns a relative path if its in the dir or an absolute path if its above the dir which causes issues in the python bundling . Removing the resolve fixes the issue, and shouldn't affect the other usages.
- Fixes https://github.com/sst/sst/issues/4114

Can delete the example if it's unnecessary just shows the example. Which is when the pyproject is a level above the python function.


thdxr:

PR #1152, opened 2024-09-26
feat: disable telemetry with env var from v2

sebws:
restore functionality from v2. I've not actually built this yet and just made the edits on Github, sorry.

I think this should be added to the documentation, but not sure if it should be added to the CLI description of telemetry?


jayair:
Thanks.

Can we test this?

Also, lets open an issue for this.

sebws:
by test, do you mean write tests or just make sure it builds etc.?

I've updated to include DO_NOT_TRACK and explicitly check the variables == 1. I've taken the language from the SST v2 docs for the description

jayair:
Yeah can we also open an issue for this?

jayair:
Oh I see it.

PR #1155, opened 2024-09-27
Update documentation to warn people about Secret naming requirements

subssn21:
The documentation does not mention the naming requirements for secrets, updating the documentation to reflect the error message you receive with an improperly named secret

### DIFF
--- a/platform/src/components/secret.ts
+++ b/platform/src/components/secret.ts
@@ -18,6 +18,10 @@ export class SecretMissingError extends VisibleError {
  * Secrets are encrypted when they are stored in your state file or in a function package.
  * :::
  *
+ * :::note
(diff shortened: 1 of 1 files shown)


jayair:
Was this an error you got on deploy?

subssn21:
Yes, that was literally the error message you get, but it's not mentioned in the documentation that their are any special requirements for secret names, so I assumed if it was a valid variable name it would work which is not the case.

jayair:
Yeah let me add it to the docs. The secret names follow the same rules as the components.

nurulhudaapon:
@jayair What was that PR about? Underscore in secret name works now.


https://github.com/sst/ion/pull/1073

PR #1191, opened 2024-10-01
fix: error on creating Cognito authorizers for V1 AWS APIs

### LINKED ISSUE #4058: Unable to reference an existing authorizer function for ApiGatewayV1
Currently I don't see anyway to reference an existing function when adding a new authorizer to `ApiGatewayV1`. 
I tried to pass the `arn` instead of the handler path without any success.
Another attempt I had was to use the `userPools` method (because I saw it's the only case where SST doesn't create a lambda) and use the transform function to update `authorizerUri`. But it ended up with the following error :
```
TypeError: Cannot read properties of undefined (reading 'nodes')
       at createAuthorizer (file:///Volumes/sredata/dev/prismicio/obelix/.sst/platform/src/components/aws/apigatewayv1-authorizer.ts:130:32)
       at new ApiGatewayV1Authorizer (file:///Volumes/sredata/dev/prismicio/obelix/.sst/platform/src/components/aws/apigatewayv1-authorizer.ts:64:24)
       at ApiGatewayV1.addAuthorizer (file:///Volumes/sredata/dev/prismicio/obelix/.sst/platform/src/components/aws/apigatewayv1.ts:1029:12)
       at services/document-search/infra/search.ts (file:///Volumes/sredata/dev/prismicio/obelix/services/document-search/infra/search.ts:26:24)
       at services/document-search/infra/mod.ts (file:///Volumes/sredata/dev/prismicio/obelix/services/document-search/infra/mod.ts:6:1)
       at run (file:///Volumes/sredata/dev/prismicio/obelix/sst.config.ts:33:31)
       at run (file:///Volumes/sredata/dev/prismicio/obelix/.sst/platform/src/auto/run.ts:32:20)
       at file:///Volumes/sredata/dev/prismicio/obelix/eval.ts:4:22
```

Finally I came up with following solution:

```typescript
const authorizer = api.addAuthorizer({
	name: "custom-authorizer",
	ttl: 0,
	identitySource: "",
	requestFunction: {
		handler: "services/dummy-authorizer/src/index.handler",
	},
	transform: {
		authorizer: {
			authorizerUri: authorizerFunction.invokeArn,
		},
	},
})
```

In this solution I let SST creating the lambda even though I don't use it.
Is there anything I'm missing here?
Also I'm pretty sure I spot a bug with `userPools` as well.

greenlynx:
Fixes the error discussed in https://github.com/sst/sst/issues/4058


jayair:
Thanks

jamesleech:
I've got this same issue in V1, 

This is causing an issue when trying to create a Cognito Authoriser.
When setting up a Cognito Authoriser we do not required a function, and in fact errors before this line if you try to add a token/request Function.

fn? will be rightly undefined in this case, and should not be force to be non null.

If I revert this change, and do the below, my authoriser is setup correctly and works.
```
api.addAuthorizer({
    name: "my-authorizer",
    userPools: [myUserPoolArn],
  });
```

jamesleech:
it does look like this issue could also be in V2 (http) of the apigateway function also.

greenlynx:
I haven't tested it, but I don't think this is an issue in V2 API Gateway - in there, as the code currently stands, it looks like `fn` will only be undefined if `lamb` is undefined, and the `fn!` line is only hit if `lamb` is truthy. I think there's an argument to be made for avoiding the use of the non-null assertion operator as far as possible, to avoid bugs creeping in inadvertantly in future if the logic changes - but I'd rather keep this PR focused on fixing the specific bug we're hitting.

Maintainers - are you happy with this PR? Do I need to do anything further to get it merged?

jamesleech:
> I haven't tested it, but I don't think this is an issue in V2 API Gateway - in there, as the code currently stands, it looks like `fn` will only be undefined if `lamb` is undefined, and the `fn!` line is only hit if `lamb` is truthy. I think there's an argument to be made for avoiding the use of the non-null assertion operator as far as possible, to avoid bugs creeping in inadvertantly in future if the logic changes - but I'd rather keep this PR focused on fixing the specific bug we're hitting.
> 
> Maintainers - are you happy with this PR? Do I need to do anything further to get it merged?


"argument to be made for avoiding the use of the non-null assertion operator as far as possible" 100%  Agree.

Would appreciate this PR being released, as I've currently out of scoped auth for a specific API route because of this bug for a project I'm working on to carry on to meet project deadlines, we'll loop back once this is merged and released.

greenlynx:
> Would appreciate this PR being released, as I've currently out of scoped auth for a specific API route because of this bug for a project I'm working on to carry on to meet project deadlines, we'll loop back once this is merged and released.

Me too, as it's blocking me as well. I don't know how to proceed though - I've never contributed to this project before and am not sure of the process.

fwang:
Will be in v3.2.29 - sorry about the delay!

jamesleech:
thanks @fwang ! great news

PR #1202, opened 2024-10-02
fix: container environment variables

### LINKED ISSUE #4012: Cluster.addService when using containers don't set environment variables
When specifying containers, the environment variables are being ignored, and the workaround of passing environment variables at the top level of addService is not possible due to this condition:

https://github.com/sst/ion/blob/64a591a62287089adbf0f09636dd9f9112112814/platform/src/components/aws/service.ts#L232

```ts
const service = cluster.addService("MyService", {
  public: {
    ports: [{ listen: "80/http", forward: "3000/http" }],
  },
  // can't use environment here...
  containers: [
    {
      name: "example-service",
      environment: {
        EXAMPLE: process.env.EXAMPLE // gets ignored
      },
    },
  ],
  dev: {
    command: "bun dev",
  },
});
```

sellooh:
Container variables are being ignored and top level environment can't be set when using `containers`

Since this condition doesn't allow to specify top level environment variables, there is no workaround atm: https://github.com/sst/ion/blob/64a591a62287089adbf0f09636dd9f9112112814/platform/src/components/aws/service.ts#L232

This PR adds container env variables to the task definition

closes https://github.com/sst/sst/issues/4012


jayair:
Thanks

psteinroe:
What's holding this pr back? Would love to get that fix.

sellooh:
Had conflicts with my PR and just fixed them - would also like to get this fixed, I'm using a patched version of the component for now but would love to just use SST's

fwang:
Releasing v3.2.30 with this fix - sorry about the delay

sellooh:
thank you!

PR #1235, opened 2024-10-10
fix: bundling with multiple dots in filename

maxscn:


thdxr:
we actually just got rid of builds in js - we build everything in node.go now

maxscn:
unlucky timing on my part, this should fix the issue in go as well.

PR #1253, opened 2024-10-11
fix: wait for resource creation before deploying api gateway v1

ryan-warner:
Sets `dependsOn` to `resources` when API Gateway V1 deployments are created. Prevents the following exception:
> BadRequestException: The REST API doesn't contain any methods...

We've seen this when deploying stages where the Gateway is configured with a deployment trigger. Can't think of any cases where this could cause problems?

### DIFF
--- a/platform/src/components/aws/apigatewayv1.ts
+++ b/platform/src/components/aws/apigatewayv1.ts
@@ -1270,7 +1270,7 @@ export class ApiGatewayV1 extends Component implements Link.Linkable {
               ),
             ),
           },
-          { parent },
+          { parent, dependsOn: resources },
         ),
       );
     }


jayair:
Can we also open an issue for this?

ryan-warner-foreflight:
Apologies for switching accounts there 🤦‍♂️ - but issue is open

fwang:
Leaving a note here. The root cause was that `Deployment` did not get triggered on API policy (or other properties of the API) changes. `dependsOn: resources` shouldn't be required b/c triggers relied on the applied value of resources already, so the dependency already exists.

bdewey-foreflight:
@fwang thanks for looking into this and closing. Was the root cause addressed in another change? Can you link any artifacts here for reference? Thank you!

bdewey-foreflight:
Looks like it was addressed here:
https://github.com/sst/ion/commit/7f5ac2929f2c5d34cff103bbabc06b764daef246#diff-6ecdaffb8923e896af590b41097cbb206b2577cc72acb16248fd247cf49f7054R1200

fwang:
yeah it's actually this line
https://github.com/sst/ion/commit/7f5ac2929f2c5d34cff103bbabc06b764daef246#diff-6ecdaffb8923e896af590b41097cbb206b2577cc72acb16248fd247cf49f7054R1238

Previously, the deployment did not take the API resource's property (ie. policy) into account when generating the triggers hash.

PR #1261, opened 2024-10-12
remove misleading parameter from Url in error

zvictor:
`provider.ErrBucketMissing` states that:

> The state bucket is missing, it may have been accidentally deleted. Go to https://console.aws.amazon.com/systems-manager/parameters/%252Fsst%252Fbootstrap/description?region=us-east-1&tab=Table and check if the state bucket mentioned there exists. If it doesn't you can recreate it or delete the `/sst/bootstrap` key to force recreation.

Having `region=us-east-1` in the URL does not do us any favour. By removing it, we increase the chances of sending the user to the last used region, which is more likely the right one.


fwang:
THanks @zvictor

PR #1240, opened 2024-10-10
add `mountPoints` to `Service`

zvictor:
Access to Containers' `mountPoints` is needed in order to attach volumes/filesystems.

This is part of sst/sst#4616 and [this](https://discord.com/channels/983865673656705025/1285733025547817052).

* https://docs.aws.amazon.com/AmazonECS/latest/developerguide/specify-efs-config.html
* https://docs.aws.amazon.com/AmazonECS/latest/developerguide/specify-volume-config.html

PR #1268, opened 2024-10-13
example: Surrealdb in AWS [Fargate with EFS]

zvictor:
**Note: This example makes use of `mountPoints` (#1240)**

## Surrealdb in AWS [Fargate with EFS]
_A low cost setup **not recommended for production**!_

You can use SST to deploy containers with persisting volumes, such as EFS.

Features included in this setup:
* `fck-nat` for much cheaper connectivity
* EFS for flexible persistent volumes
* Fargate Spot has preference over Fargate, reducing direct hosting costs.

---

DependsOn: #1240


examples/aws-surrealdb-fargate/sst.config.ts
```diff
@@ -0,0 +1,160 @@
...
+const SURREAL_PASSWORD = "root";
+const SURREAL_NAMESPACE = "test";
+const SURREAL_DATABASE = "test";
+
+// Network
+const CONTAINER_PORT = 8080;
+
+/**
+ * ## Surrealdb in AWS [Fargate with EFS]
+ * _A low cost setup **not recommended for production!_
```
zvictor:
```suggestion
 * _A low cost setup **not recommended for production**!_
```

fwang:
Thanks @zvictor!

I updated it to use the new EFS component. Also left out the spot instance part, that can be a separate PR.

PR #1271, opened 2024-10-13
chore: update open-next version in Nextjs component 

sommeeeer:
This PR bumps the default OpenNext version to be the latest which is `3.1.4`. It also changes the build command to address that OpenNext => `3.1.13` needs to use the new package called: `@opennextjs/aws`. 

If its `3.1.13` or lower it will use the old `open-next` package.

### DIFF
--- a/platform/src/util/compare-semver.ts
+++ b/platform/src/util/compare-semver.ts
@@ -1,0 +1,19 @@
+export function compareSemver(v1: string, v2: string): number {
+  if (v1 === "latest") return 1;
+  if (/^[^\d]/.test(v1)) {
+    v1 = v1.substring(1);
+  }
+  if (/^[^\d]/.test(v2)) {
+    v2 = v2.substring(1);
+  }
+  const [major1, minor1, patch1] = v1.split(".").map(Number);
+  const [major2, minor2, patch2] = v2.split(".").map(Number);
+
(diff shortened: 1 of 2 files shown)


conico974 (approved):
LGTM

platform/src/util/compare-semver.ts
```diff
@@ -0,0 +1,19 @@
...
+  const [major1, minor1, patch1] = v1.split(".").map(Number);
+  const [major2, minor2, patch2] = v2.split(".").map(Number);
+
+  if (major1 !== major2) return major1 - major2;
+  if (minor1 !== minor2) return minor1 - minor2;
+  return patch1 - patch2;
+}
+
+export function getOpenNextPackage(openNextVersion: string): string {
+  return compareSemver(openNextVersion, "3.1.13") <= 0 ? "open-next" : "@opennextjs/aws";
```
socsieng, line 18: Is this a typo? Should it be `3.1.3`?

conico974: Ho yeah definitely a typo @sommeeeer @thdxr

sommeeeer: > Is this a typo? Should it be `3.1.3`?

nice spotted! making a new PR to address this now.

PR #1286, opened 2024-10-16
feat: add Deno format ignore directive to generated files

zxt-tzx:
See https://docs.deno.com/runtime/reference/cli/formatter/


jayair:
Thanks

PR #1290, opened 2024-10-16
Fix user pool triggers type

tropicbliss:
Currently:

```ts
const kmsKey = new aws.kms.Key("EmailSenderKey");

export const userPool = new sst.aws.CognitoUserPool("UserPool", {
  triggers: {
    kmsKey: kmsKey.arn as any,
    ...
  },
  ...
});
```


fwang:
Thanks @tropicbliss!

ryanleecode:
this was reported here https://github.com/sst/sst/issues/3972

PR #1294, opened 2024-10-16
Fix sdk event tree shake and import

zohaibakber:


zohaibakber:
Further testing this, seems like it still have issues, working on this, Dax can close this if he happens to fix it first.

thdxr:

PR #1322, opened 2024-10-20
Provide the ability to specify a response headers policy

JohnG45:
We are having to do this currently, which is pretty nasty:
```
new sst.aws.Nextjs('app', {
  ...args,
  transform: {
    cdn: {
      transform: {
        distribution: (args) => {
          args.defaultCacheBehavior = {
            ...args.defaultCacheBehavior,
            responseHeadersPolicyId: responseHeadersPolicy,
          }
          args.orderedCacheBehaviors = args.orderedCacheBehaviors
            ? $output(args.orderedCacheBehaviors).apply((behaviours) =>
                behaviours.map((behaviour) => ({ ...behaviour, responseHeadersPolicyId: responseHeadersPolicy }))
              )
            : undefined
        },
      },
    },
  },
})
```
After these changes:
```
new sst.aws.Nextjs('app', {
  ...args,
  responseHeadersPolicy
})
```
There was a `responseHeadersPolicy` prop in SST v2, and we were using it, so it's likely that others are too.

### DIFF
--- a/platform/src/components/aws/analog.ts
+++ b/platform/src/components/aws/analog.ts
@@ -269,6 +269,17 @@ export interface AnalogArgs extends SsrSiteArgs {
    * ```
    */
   cachePolicy?: SsrSiteArgs["cachePolicy"];
+  /**
+   * Configure the Analog app to use an existing CloudFront response headers policy.
+   *
+   * @example
+   * ```js
+   * {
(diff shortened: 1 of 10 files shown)


fwang:
@JohnG45 we're in the middle of switching repos (unfortunately github is making this quite messy) - going to close this PR for now. I have noted this down. And will get to it once we've moved over to sst/sst

